PR #12, opened 2015-01-16
Added Feed API and tests.

alexgerstein:


app/api/feeds.py
```diff
@@ -0,0 +1,71 @@
...
+    'url': fields.String,
+    'last_updated': fields.DateTime,
+    'uri': fields.Url('feed')
+}
+
+class FeedListAPI(Resource):
+
+	def __init__(self):
+		self.reqparse = reqparse.RequestParser()
+		self.reqparse.add_argument('title', type=str, required=True,
```
swglad: Is 'title' strictly required or something we should fill in given the URL since we won't be scraping at first?

app/api/feeds.py
```diff
@@ -0,0 +1,71 @@
...
+
+	def post(self):
+		args = self.reqparse.parse_args()
+		feed = {
+			'id': feeds[-1]['id'] + 1,
+			'title': args['title'],
+			'url': args['url'],
+			'last_updated': datetime.min
+		}
+		feeds.append(feed)
```
swglad: where do you initialize this 'feeds' list?

app/api/feeds.py
```diff
@@ -0,0 +1,71 @@
...
+		return {'feed': marshal(feed, feed_fields)}, 201
+
+class FeedAPI(Resource):
+
+	def get(self, id):
+		feed = filter(lambda feed: feed['id'] == id, feeds)
+		if not feed:
+			abort(404)
+		return {'feed': marshal(feed[0], feed_fields)}
```
swglad: What's feed[0]?

app/api/feeds.py
alexgerstein: Added feedparser to extract title from URL.

app/api/feeds.py
alexgerstein: Filter() returns an array. Need to get just the first instance.

PR #13, opened 2015-01-16
built out user api and tests

swglad:


app/__init__.py
```diff
@@ -10,4 +9,7 @@
     stream_handler = logging.StreamHandler()
     app.logger.addHandler(stream_handler)
     app.logger.setLevel(logging.INFO)
-    app.logger.info('rss-aggregator startup')
\ No newline at end of file
+    app.logger.info('rss-aggregator startup')
+
+from app.frontend import dashboard
```
alexgerstein: Move this back to top. This is more important that Logging.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+		'description': ""
+	}
+]
+
+user_fields = {
+	'first_name': fields.String,
+	'last_name': fields.String,
+	'screenname': fields.String,
+	'email': fields.String,
+	'location': fields.String(default = ""),
```
alexgerstein: Maybe stick to vital fields for now? So no location or description.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+}
+
+class UserListAPI(Resource):
+	def __init__(self):
+		self.reqparse = reqparse.RequestParser()
+		self.reqparse.add_argument('first_name', type = str, required = True)
+		self.reqparse.add_argument('last_name', type = str, required = True)
+		self.reqparse.add_argument('screenname', type = str, required = True)
+		self.reqparse.add_argument('email', type = str, required = True)
+		self.reqparse.add_argument('location', type = str, required = False)
```
alexgerstein: I think required defaults to false, so you can just not include it when it is.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+		self.reqparse.add_argument('description', type = str, default = "", help = 'Something about you')
+		super(UserListAPI, self).__init__()
+
+	def get(self):
+		''' Show all Users '''
```
alexgerstein: Want to keep the comments? Or do you think we can take them out?

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+	# 	}),
+	'uri': fields.Url('user')
+}
+
+class UserListAPI(Resource):
+	def __init__(self):
+		self.reqparse = reqparse.RequestParser()
+		self.reqparse.add_argument('first_name', type = str, required = True)
+		self.reqparse.add_argument('last_name', type = str, required = True)
+		self.reqparse.add_argument('screenname', type = str, required = True)
```
alexgerstein: I think we should just use email as username/screename. An RSS reader doesn't have many public aspects so we don't need this or a description. We arguably don't even need first and last name either.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+		for key, val in args.iteritems():
+			if val != None:
+				user[key] = val
+		user['id'] = user_list[-1]['id'] + 1
+		return { 'user': user}, 201
+
+class UserAPI(Resource):
+	def __init__(self):
+		self.reqparse = reqparse.RequestParser()
+		self.reqparse.add_argument('first_name', type = str)
```
alexgerstein: See comments above about fields.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+		self.reqparse.add_argument('last_name', type = str)
+		self.reqparse.add_argument('screenname', type = str)
+		self.reqparse.add_argument('email', type = str)
+		self.reqparse.add_argument('location', type = str)
+		self.reqparse.add_argument('description', type = str, help = 'Something about you')
+		super(UserAPI, self).__init__()
+
+	def get(self, id):
+		''' Show User @id '''
+		return { 'user': [marshal(user, user_fields) for user in user_list if user['id']==id] }
```
alexgerstein: 'user' should map to a one user representation, not a list.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+
+	def put(self, id):
+		''' Edit User @id '''
+		args = self.reqparse.parse_args()
+		user = {}
```
alexgerstein: This should be set to the current user in the database so that if the put request doesn't include all reqparse arguments, it doesn't remove them either.

app/api/users.py
```diff
@@ -0,0 +1,95 @@
...
+		for key, val in args.iteritems():
+			if val != None:
+				user[key] = val
+		return { 'user': user }
+
+	def delete(self, id):
+		''' Destroy User @id '''
```
alexgerstein: You can write the delete. See how I tested it on my pull request.

tests/api_tests/user_tests.py
```diff
@@ -0,0 +1,27 @@
...
+		self.app = app.test_client()
+
+	def tearDown(self):
+		pass
+
+	def test_get(self):
+		rv = self.app.get('/api/users')
+		eq_(len(rv.data),355) # size of sample input text
+
+	def test_post(self):
```
alexgerstein: More tests! There should be ones to make sure the required fields are there. What should happen if email isn't present?

What about puts? gets?

app/api/users.py
swglad: Can we leave them for now? It's still helpful to me

alexgerstein: Sure. No problemo.

app/__init__.py
swglad: Doesn't seem to work if I move the import before 'app = Flask'

alexgerstein: After that, before logging.

app/api/users.py
```diff
@@ -0,0 +1,64 @@
...
+		super(UserAPI, self).__init__()
+
+	def get(self, id):
+		''' Show User @id '''
+		return { 'user': [marshal(user, user_fields) for user in user_list if user['id']==id][0] }
+
+	def put(self, id):
+		''' Edit User @id '''
+		args = self.reqparse.parse_args()
+		user = [ user for user in user_list if user['id']==id ][0]
```
swglad: feed = filter(lambda feed: feed['id'] == id, feeds)

app/api/users.py
```diff
@@ -0,0 +1,64 @@
...
+
+class UserAPI(Resource):
+	def __init__(self):
+		self.reqparse = reqparse.RequestParser()
+		self.reqparse.add_argument('email', type = str, required = True)
+		super(UserAPI, self).__init__()
+
+	def get(self, id):
+		''' Show User @id '''
+		return { 'user': [marshal(user, user_fields) for user in user_list if user['id']==id][0] }
```
swglad: feed = filter(lambda feed: feed['id'] == id, feeds)

PR #19, opened 2015-01-18
User api database

swglad:


app/api/users.py
```diff
@@ -52,30 +36,29 @@ def __init__(self):
...
 		args = self.reqparse.parse_args()
-		try:
-			user = next(get_users_by_id(id))
-		except StopIteration:
+		user = User.query.get(id)
+		if not user:
 			abort(404)
-		user['email'] = args.email
+		user.email = args.email
+		db.session.add(user)
```
alexgerstein: I don't think you need to have the db.session.add line. See: http://stackoverflow.com/questions/11292612/flask-sqlalchemy-how-to-conditionally-insert-or-update-a-row

PR #20, opened 2015-01-19
Convert feed API to use database.

alexgerstein:


.travis.yml
```diff
@@ -8,4 +8,6 @@ deploy:
   provider: heroku
   api_key:
     secure: OeO43wvD0lvzZu2JL+/nqfSOvYjAfNrSK/qFi2BWWBe0kGKNRiFHmvHgJ7bypTzI+0+25tGS4uanhIf8ZR5gToxSLi5FymOwD9LzRHqyoxxPbpnnclUpIRT5ia6x/VHhnnKzdb4hiE7GUHrIKcbWO3mdIvJljhni4ipNy7D/N1w=
-  app: peaceful-falls-6831
\ No newline at end of file
+  app: peaceful-falls-6831
+after_deploy:
+  - ./db_upgrade.py
```
swglad: why db_upgrade and not db_migration? I believe that it's correct, just can't remember the explanation

app/api/feeds.py
```diff
@@ -39,6 +26,8 @@ def __init__(self):
 		super(FeedListAPI, self).__init__()
 
 	def get(self):
+		feeds = Feed.query.all()
```
swglad: Make it a one-liner?

PR #21, opened 2015-01-19
Add config files to simplify dev process.

alexgerstein:
This way secret keys will stay secret.

All API keys should be stored in instance/config.py on local machine and in Heroku environment variables. NOT IN THE GITHUB REPO.

Also, note that with this change, tests should be run locally with ./test.sh, not nosetests directly. And similarly, you should use ./start.sh, instead of run.py directly.


alexgerstein:
Closes #10

PR #29, opened 2015-01-25
Added subscribe/unsubscribe to API.

### LINKED ISSUE #15: Update API for database requests

alexgerstein:
Closes #15.


alexgerstein:
This will fix the Heroku bug that has it offline, along with closing #15.

app/api/subscriptions.py
```diff
@@ -43,23 +20,27 @@ def __init__(self):
...
             abort(400)
 
-        subscription = {
-            'user_id': logged_in_user_id,
-            'feed_id': feed_id,
-            'mergeable': True
-        }
-        subscriptions.append(subscription)
+        if not feed:
+            abort(400)
```
swglad: Do you need try/except AND if not feed?

alexgerstein: Yeah. The try is in case
```Python 
args['feed_uri'].split('/')[-1]
```
is not an int. If the id simply doesn't exist in the DB, then it will return None instead of raising an exception.

I will specify the exception to catch on ValueError.

tests/api_tests/user_tests.py
```diff
@@ -14,10 +14,10 @@ def setUp(self):
         self.app = app.test_client()
         db.create_all()
         self.user_role = RoleFactory()
-        user = UserFactory.create(role=self.user_role)
+        user = UserFactory.create(roles=[self.user_role])
```
swglad: For consistency, in subscription tests, you used: self.user = UserFactory(roles=[role],subscribed=[feed]).

Let's either remove the .create or add it to the others.

tests/model_tests/user_tests.py
```diff
@@ -13,7 +13,7 @@ def setUp(self):
         self.app = app.test_client()
         db.create_all()
         user_role = RoleFactory()
-        user = UserFactory.create(role=user_role)
+        user = UserFactory.create(roles=[user_role])
```
swglad: one more

PR #42, opened 2015-02-01
Visualize feed

swglad:


app/frontend/dashboards.py
```diff
@@ -1,6 +1,6 @@
 from app import app
 
-from flask import abort, render_template, redirect
+from flask import abort, render_template, redirect, url_for
```
alexgerstein: Don't need url_for anymore.

alexgerstein:
This also now integrates functional tests through sauce labs and should resolve the heroku create error.

PR #64, opened 2015-02-06
remove POST from feeds (now in subscriptions)

### LINKED ISSUE #49: Do we need a POST for feeds?
The only time a feed will get added to our DB is when a user tries to subscribe, and right now that post is happening as a 'subscriptions' POST.

So should we maybe delete POST for feeds?

swglad:
resolves #49

### DIFF
--- a/sourcemash/api/feeds.py
+++ b/sourcemash/api/feeds.py
@@ -19,27 +19,10 @@ feed_fields = {
 class FeedListAPI(Resource):
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


sourcemash/api/feeds.py
```diff
@@ -19,28 +19,11 @@
 class FeedListAPI(Resource):
 
     def __init__(self):
```
alexgerstein: We don't need the init at all anymore.

PR #65, opened 2015-02-07
Add full text to scraper.

### LINKED ISSUE #63: Store full-text of articles.
In order to analyze articles, we need to have the full body text; we can't simply rely on the excerpts/descriptions.

I think the most time-sensitive, logical approach is to use https://github.com/buriy/python-readability. Although this may feel like cheating, maybe we'll find some ways to improve their code and contribute to a real open-source project!

alexgerstein:
Fixes #63.


sourcemash/frontend/frontend.py
```diff
@@ -31,7 +31,7 @@ def feeds():
 def feed(id):
 	''' Get articles & pass through to render_template '''
 	feed = Feed.query.get(id)
-	feed_items = Item.query.filter(Item.feed_id==id).all()
+	feed_items = Item.query.filter(feed_id==id).all()
```
swglad: Where is feed_id defined here?

sourcemash/frontend/templates/feed.html
```diff
@@ -15,7 +15,7 @@
 	            <div class="card-content black-text">
 	              <span class="card-title"><h4><a href="{{ item.link }}">{{ item.title }}</a></h4></span>
 	              	<p>{{ item.author }}</p>
-					<p>{{ item.summary }}</p>
+					<p>{{ item.text }}</p>
```
swglad: Do we really want to be putting the full text here and not the summary?

sourcemash/items/models.py
```diff
@@ -7,6 +7,7 @@ class Item(db.Model):
     last_updated = db.Column(db.DateTime)
     author = db.Column(db.String(120))
     summary = db.Column(db.String(500))
+    text = db.Column(db.Text)
```
swglad: Are there any length requirements on this that we should keep in mind? Will this need to make us worry about database row issues?

sourcemash/frontend/frontend.py
alexgerstein: Good catch. I forgot to switch it to filter_by.

sourcemash/frontend/templates/feed.html
alexgerstein: We do want the summary, but I didn't want to make a new page for now since you'll be working on the backend changes.

So to make sure full text was being generated properly, I just put it on the feed page for now.

worker_tasks/feed_scraper.py
```diff
@@ -17,12 +19,16 @@
 celery = Celery(app.name, broker=app.config['CELERY_BROKER_URL'])
 celery.conf.update(app.config)
 
+
+def get_full_text(url):
+    html = requests.get(url).content
+    return Document(html).summary()
```
swglad: Can you link to documentation for this? Trying to find it online and having trouble.

For example, what does .summary() do and why isn't it something like .get_text() ?

sourcemash/items/models.py
alexgerstein: Text doesn't take a min length; it will be too tough to pick a max length.

See: http://docs.sqlalchemy.org/en/rel_0_9/core/type_basics.html#sqlalchemy.types.Text

sourcemash/frontend/frontend.py
```diff
@@ -31,7 +31,7 @@ def feeds():
 def feed(id):
 	''' Get articles & pass through to render_template '''
 	feed = Feed.query.get(id)
-	feed_items = Item.query.filter(Item.feed_id==id).all()
+	feed_items = Item.query.filter_by(feed_id=id).all()
```
swglad: How is feed_id declared here?

sourcemash/frontend/templates/feed.html
swglad: Alright.

worker_tasks/feed_scraper.py
alexgerstein: You can look at the source code for summary here: https://github.com/buriy/python-readability/blob/0c2f29ed0df927e4305331f7411a5a1efe2c9cbc/readability/readability.py

It's what python-readability calls it's main function which gets the best candidate for the article and parses it down to the relevant tags.

sourcemash/frontend/frontend.py
alexgerstein: Feed_id is the name of a parameter to filter_by, so it doesn't need to be declared, it just looks up the column in the Item table.

See: http://docs.sqlalchemy.org/en/rel_0_9/orm/query.html#sqlalchemy.orm.query.Query.filter_by

PR #80, opened 2015-02-11
Categorization algorithm

### LINKED ISSUE #71: Categorization Algorithm
Can we simply rely on proper nouns? What if the categorization goes deeper?

__We should write this separate of our site so that we can test our assumptions first.__

swglad:
Resolves #71 Categorization Algorithm


sourcemash/categorize.py
```diff
@@ -0,0 +1,28 @@
...
+	cat1 = ""
+	cat2 = ""
+
+	categories = Counter()
+
+	title = title.split()
+	text = text.split()
+
+	# Get word counts from title, text of article
+	for source in [title, text]:
```
alexgerstein: This might be cleaner if you iterated over (title + text). Then you won't need the nested for loop on line 14.

sourcemash/categorize.py
```diff
@@ -0,0 +1,28 @@
+from collections import Counter
+
+def categorizeItem(title, text, category_dict):
+	cat1 = ""
+	cat2 = ""
+
+	categories = Counter()
+
+	title = title.split()
```
alexgerstein: We only care if in category_dict() for both title and text words.

PR #83, opened 2015-02-12
db and api integration for feed-item scraper

### LINKED ISSUE #47: Create db seed file.
It would be nice if we had a command for the DB to populate itself with users, feeds, etc whenever we create a db for the first time (i.e. after rm app.db, or during testing).

It probably makes the most sense to do this as a manage.py script. For example,

```python
@manager.command
def seed():
    "Add seed data to the database."
    db.session.add(...)
    db.session.commit()
```

### LINKED ISSUE #82: Scraper should update feed-items db with categories
Finish scraper look to do database updates/commits for categories.

swglad:
Resolves #82 categorization with db, api integration
Closes #47 seed database --> required: need to delete app.db & run db-upgrade first


alexgerstein:
Rebase with master to fix up the migrations folder. Then I'll do a quick review.

sourcemash/categorize.py
```diff
@@ -2,6 +2,8 @@
 from string import punctuation
 
 def categorizeItem(title, text, category_dict):
```
alexgerstein: Make this categorize_item.

sourcemash/categorize.py
```diff
@@ -20,4 +22,9 @@ def categorizeItem(title, text, category_dict):
 	# Return top 2 words in categories
 	most_frequent = categories.most_common(2)
 
-	return (most_frequent[0][0], most_frequent[1][0])
\ No newline at end of file
+	if len(most_frequent) > 0:
```
alexgerstein: Can you write a test for this part? Just so we don't change it in the future with unexpected consequences.

i.e. set text to "the of" and make sure "" and "" are returned.

worker_tasks/scraper.py
```diff
@@ -74,6 +77,10 @@ def scrape_articles(self):
         for feed in Feed.query.all():
             self.store_items_and_category_counts(feed)
 
-        # TODO: Assign categories to articles
-
-
+        # Assign categories to articles after all feeds are processed
+        for item in Item.query.filter_by(category_1=None).all():
+            (cat1, cat2) = categorizeItem(item.title, item.text, self.title_categories)
```
alexgerstein: you'll have to rename this here too to 'categorize_item'

worker_tasks/scraper.py
```diff
@@ -74,6 +77,10 @@ def scrape_articles(self):
...
+            item.category_1 = cat1
+            item.category_2 = cat2
+            db.session.commit()
+            logging.info("Categorized item %s: %s, %s" % (item.title, item.category_1, item.category_2))
```
alexgerstein: Just to make logging easier to read among the stream of text, can we rewrite this to something like:

CATEGORIZED [%s]: (%s, %s)

sourcemash/categorize.py
```diff
@@ -1,7 +1,9 @@
 from collections import Counter
 from string import punctuation
 
-def categorizeItem(title, text, category_dict):
+def categorize_item(title, text, category_dict):
```
alexgerstein: Would it make sense to move this to the scraper? It feels weird passing around our category_dict.

tests/test_worker/test_categorization.py
```diff
@@ -1,13 +1,23 @@
...
+        category_dict.update(ebolaItem.title.split())
+        
+        (cat1, cat2) = categorize_item(ebolaItem.title, ebolaItem.text, category_dict)
 
-		assert (cat1, cat2) == ("Ebola", "West")
\ No newline at end of file
+        assert (cat1, cat2) == ("Ebola", "West")
+
+    def test_empty_categories(sef, oftheItem):
+        category_dict = Counter()
```
alexgerstein: Instead of manually making the category dict in the test, you should probably create a scraper class and run the reset dictionary function on it.

tests/test_worker/test_categorization.py
```diff
@@ -1,13 +1,23 @@
...
+        category_dict.update(ebolaItem.title.split())
+        
+        (cat1, cat2) = categorize_item(ebolaItem.title, ebolaItem.text, category_dict)
 
-		assert (cat1, cat2) == ("Ebola", "West")
\ No newline at end of file
+        assert (cat1, cat2) == ("Ebola", "West")
+
+    def test_empty_categories(sef, oftheItem):
```
alexgerstein: Typo: "Sef" should be "self"

alexgerstein: I wouldn't make a factory/fixture for the of/the item. Instead just pass a custom title and text to categorize_item.

tests/fixtures/item_fixtures.py
```diff
@@ -7,4 +7,8 @@ def item(request):
 
 @pytest.fixture()
 def ebolaItem(request):
-	return item_factories.EbolaItemFactory()
\ No newline at end of file
+	return item_factories.EbolaItemFactory()
+
+@pytest.fixture()
```
alexgerstein: Let's not make this a fixture/factory.

tests/fixtures/app_fixtures.py
```diff
@@ -50,6 +50,7 @@ def session(db, request):
     user_factories.UserFactory._meta.sqlalchemy_session = session
     item_factories.ItemFactory._meta.sqlalchemy_session = session
     item_factories.EbolaItemFactory._meta.sqlalchemy_session = session
+    item_factories.OfTheItemFactory._meta.sqlalchemy_session = session
```
alexgerstein: Nah...

tests/factories/item_factories.py
```diff
@@ -20,4 +20,8 @@ class EbolaItemFactory(ItemFactory):
...
+
+class OfTheItemFactory(ItemFactory):
```
alexgerstein: This feels silly since we'll only use it for the one test.

alexgerstein: Maybe we want to think about making a factory for a scraper instead?

sourcemash/categorize.py
swglad: No. The categorization algorithm is its own unit that exists distinctly from the scraper, and it's precondition is string article title, string article text, and dictionary/list of possible categories.

sourcemash/categorize.py
swglad: Done.

sourcemash/categorize.py
alexgerstein: But category_dict exists for the sole purpose of categorize_item, so it's weird for it to be passed in from scraper. Maybe scraper shouldn't be the class and we should extract the "class" stuff (i.e. the category_dict) to a Categorizer class.

alexgerstein:
Big changes, but I think we should do the following:

* The scraper shouldn’t be the class since it’s sequential and we can’t really test it; a Categorizer class should be instead

* Categorizer will have:
    * reset_title_categories
    * parse_title_categories
    * categorize item
    * helper to determine if a word should "count" in the dict
Note: the inputs for these functions only need to be strings of text.

The scraper will deal with actually accumulating these strings from either an RSS feed or the DB, and so it does not need to be a class.

tests/factories/item_factories.py
```diff
@@ -14,10 +14,4 @@ class ItemFactory(SQLAlchemyModelFactory):
     author = 'Scott Gladstone'
     summary = 'Summary of the feed item.'
     text = 'Text of the feed item.'
-    feed = factory.SubFactory(FeedFactory)
-
-class EbolaItemFactory(ItemFactory):
```
alexgerstein: I think EbolaItem can stay as a factory. It was the other one that was weird.

swglad: There's still only one test for it, so I don't think it's necessary for now. If we end up determining that we want multiple tests using EbolaItem, we'll add it back later.

requirements.txt
```diff
@@ -50,7 +47,11 @@ requests==2.5.1
 sauceclient==0.1.0
 selenium==2.44.0
 six==1.9.0
+SQLAlchemy==0.9.8
 sqlalchemy-migrate==0.9.4
 sqlparse==0.1.14
+Tempita==0.5.2
```
alexgerstein: Do we need this? Can you do a pip uninstall and refreeze?

requirements.txt
```diff
@@ -29,11 +11,26 @@ enum34==1.0.4
...
+Flask-WTF==0.10.3
 gunicorn==19.1.1
 itsdangerous==0.24
+Jinja2==2.7.3
 jsmin==2.1.0
 kombu==3.0.24
 lxml==2.3.6
+Mako==1.0.1
+MarkupSafe==0.23
+nose==1.3.4
```
alexgerstein: Do we need this? Can you do a pip uninstall and refreeze?
Add a line note

requirements.txt
```diff
@@ -29,11 +11,26 @@ enum34==1.0.4
...
+Flask-Security==1.7.4
+Flask-SQLAlchemy==2.0
+Flask-WTF==0.10.3
 gunicorn==19.1.1
 itsdangerous==0.24
+Jinja2==2.7.3
 jsmin==2.1.0
 kombu==3.0.24
 lxml==2.3.6
+Mako==1.0.1
```
alexgerstein: Do we need this? Can you do a pip uninstall and refreeze?
Add a line note

sourcemash/categorize.py
```diff
@@ -1,23 +1,99 @@
...
+        text = filter(lambda word: word in self.title_categories, text.split())
+        title = [word.strip(punctuation) for word in title]
+        text = [word.strip(punctuation) for word in text]
+
+        # Get word counts from title, text of article
+        categories.update(title + text)
+
+        # Give weights to categories found in title
+        for word in title:
+            categories[word] *= 2
```
alexgerstein: Extract '2' out to CONSTANT, like "TITLE_WORD_WEIGHT," or something of that nature.

worker_tasks/scraper.py
```diff
@@ -13,67 +12,57 @@
 
 import logging
 
+categorizer = Categorizer()
```
alexgerstein: This should be declared in manage.py.

worker_tasks/scraper.py
```diff
@@ -13,67 +12,57 @@
...
+        item.category_1 = cat1
+        item.category_2 = cat2
+        db.session.commit()
+        logging.info("CATEGORIZED [%s]: (%s, %s)" % (item.title, item.category_1, item.category_2))
 
-    def parse_title_categories(self, titles):
-        for title in titles:
-            self.title_categories.update([word.strip(punctuation) for word in title.split()])
 
+def reset_title_categories():
```
alexgerstein: We shouldn't use this.

tests/factories/item_factories.py
alexgerstein: If you insist. It just makes test_categorization.py so messy, so i thought it might be nice to extract it since it's such a hefty string.

manage.py
```diff
@@ -43,19 +48,43 @@ def test(all=False):
...
+def seed():
+    """Add seed data to the database"""
+    """Required: Need to delete database & run db upgrade first"""
+    # One user
+    user = User(email="admin@sourcemash.com", password="password")
+    db.session.add(user)
+    db.session.commit()
+
+    # One feed
+    feed = Feed(id=1, title='TechCrunch > Startups',
```
alexgerstein: Maybe just use the NYTimes factory here since we already made it?

sourcemash/categorize.py
```diff
@@ -1,23 +1,101 @@
...
+
+    def parse_title_categories(self, titles):
+        for title in titles:
+            for word in title.split():
+                word = word.strip(punctuation)
+                if not self.ignore_word(word):
+                    self.title_categories.update([word])
+
+
+    def ignore_word(self, word):
```
alexgerstein: Since this is a boolean function, can we rename it to something like is_valid_word?

worker_tasks/scraper.py
```diff
@@ -1,9 +1,8 @@
-
 from sourcemash.database import db
 
 from sourcemash.models import Item, Feed
-from datetime import datetime, timedelta
-from time import mktime
+from sourcemash.categorize import Categorizer
```
alexgerstein: Do we need to import this anymore?

manage.py
swglad: Changed NYTFeedFactory to TechcrunchFeedFactory, but don't want to import factories from tests here, so we'll continue to do it manually.

PR #90, opened 2015-02-15
Transfer feeds and feedviews over to backbone.

### LINKED ISSUE #8: Convert front-end to start using Backbone.js
For a good explanation of backbone, first read the introduction in their docs and then look at the classic Todo example, http://backbonejs.org/docs/todos.html.

For a good example of how to organize backbone files, see https://github.com/kevinbeaty/flask-todomvc. The tutorial covers all of the API work that we've done, so you don't need to read the steps. This is more for folder organization later on.

This will make things much easier on the front-end.

### LINKED ISSUE #81: Visualize categories: list categories in feed-items
Resolved by #90

alexgerstein:
Closes #8. Resolves #81.


config/default.py
```diff
@@ -7,8 +7,7 @@
 
 SECRET_KEY = 'thisisnearlyimpossible'
 
-WTF_CSRF_ENABLED = True
-WTF_CSRF_SECRET_KEY = 'you-will-never-guess'
+WTF_CSRF_ENABLED = False
```
swglad: Why False for default?

sourcemash/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 def create_app(env=None):
 	app = Flask(__name__, instance_relative_config=True,
-				template_folder='frontend/templates',
+				template_folder='frontend/static/templates',
```
swglad: Why did all templates move to static/ folder? Aren't they non-static because different things can be rendered based on Backbone?

sourcemash/api/feeds.py
```diff
@@ -10,10 +10,10 @@
 from sourcemash.models import Feed
 
 feed_fields = {
+    'id': fields.Integer,
```
swglad: Why id and not uri?

sourcemash/api/subscriptions.py
```diff
@@ -10,11 +10,11 @@
...
 from datetime import datetime
 
+from feeds import feed_fields
+
 subscription_fields = {
     'id': fields.Integer,
-    'title': fields.String,
-    'url': fields.Url('frontend.feed'),
-    'uri': fields.Url('api.subscription')
+    'feed': fields.Nested(feed_fields)
```
swglad: Cool!

sourcemash/api/subscriptions.py
```diff
@@ -85,12 +93,14 @@ def put(self, id):
         if not subscription:
             abort(404)
 
+        subscription.feed = Feed.query.get(id)
+
         if args.mergeable:
```
swglad: Let's delete mergeable since we won't be using it like this anymore.

sourcemash/frontend/assets.py
```diff
@@ -12,18 +12,27 @@
 js_default = Bundle("js/default.js",
 					filters="jsmin", output="js/default.min.js")
 
-js_feeds = Bundle("js/models/feed.js",
-                    "js/collections/feeds.js",
-                    "js/views/feeds-view.js",
+
+js_templates = Bundle("templates/*.tpl",
```
swglad: What files are .tpl? I don't see those as added in this PR?

swglad:
This is way cool and an insane amount of work. Killing it.

config/default.py
alexgerstein: Because Backbone can't use WTForms, so for now I disabled CSRF. I added an issue to add CSRF back in the future, but we'll probably have to use something like this: http://danialk.github.io/blog/2013/07/28/advanced-security-in-backbone-application/

sourcemash/__init__.py
alexgerstein: Forgot I did that. I might be needed to use the JST object, but I'll try moving it back out.

sourcemash/api/feeds.py
alexgerstein: We never use the URI anymore since we define the url paths on the backbone models. Id is necessary now instead.

sourcemash/api/subscriptions.py
alexgerstein: Good call. I'll do that now.

sourcemash/frontend/templates/feed.tpl
```diff
@@ -0,0 +1,15 @@
...
+  <%= feed.get('title') %>
+</h2>
+
+<ul class="list-group">
+  <% feed.items.each(function(item) { %>
+    <div class="card-panel blue lighten-4">
+      <div class="card-content black-text">
+        <span class="card-title"><h4><a href="<%= '#/entries/' + item.get('id') %>"><%= item.get('title') %></a></h4></span>
+          <p><%= item.get('author') %></p>
+          <p><%= item.get('text') %></p>
```
swglad: Add something like: 

<p>Categories:<%= item.get('category_1') %>,<%= item.get('category_2') %></p>

sourcemash/api/items.py
```diff
@@ -13,7 +14,6 @@
     'author': fields.String,
     'category': fields.String,
```
swglad: 'category_1': fields.String,
    'category_2': fields.String,

sourcemash/frontend/templates/feed.tpl
```diff
@@ -0,0 +1,15 @@
+<h2>
+  <%= feed.get('title') %>
+</h2>
+
+<ul class="list-group">
+  <% feed.items.each(function(item) { %>
+    <div class="card-panel blue lighten-4">
+      <div class="card-content black-text">
+        <span class="card-title"><h4><a href="<%= '#/entries/' + item.get('id') %>"><%= item.get('title') %></a></h4></span>
```
swglad: How does this #/entries work?

alexgerstein: It doesn't. I'll fix it.

PR #118, opened 2015-02-18
Voting-api-db

### LINKED ISSUE #75: Upvote/Downvote for articles (Backend)
API: 
- 'PUT' request on feed-item (api/feeds/'<id>'/items/'<id>') for upvote/downvote

DB: 
1. 'items' table needs a new column "totalVotes"
2. 'votes' (named: user-item) table with user_id-item_id-VoteUp(True if +1, False if -1)
-----> Resolves Issue of keeping track of whether a user has already voted, by doing a lookup to this table
-----> Also allows a "filter by upvote" feature on front-end where we can return all articles that they've upvoted. Similar to "Starred"/"Pinned"/"Favorited"/etc.
------> Helps keep track of unread articles if we add an extra boolean column Unread later

swglad:
Closes #75


sourcemash/api/items.py
```diff
@@ -15,8 +16,7 @@
     'category_1': fields.String,
     'category_2': fields.String,
     'summary': fields.String,
-    'category_1': fields.String,
-    'category_2': fields.String,
```
swglad: category_1 and category_2 snuck in their twice somehow.

migrations/versions/426f1ad50ca2_.py
```diff
@@ -0,0 +1,34 @@
...
+revision = '426f1ad50ca2'
+down_revision = '598562686e8'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('items_users',
```
alexgerstein: Let's name this user_items.

sourcemash/items/models.py
```diff
@@ -11,7 +11,7 @@ class Item(db.Model):
     feed_id = db.Column(db.Integer, db.ForeignKey('feed.id'))
     category_1 = db.Column(db.String(100))
     category_2 = db.Column(db.String(100))
-
+    totalVotes = db.Column(db.Integer)
```
alexgerstein: Set default here to 0.

tests/factories/item_factories.py
```diff
@@ -15,6 +15,7 @@ class ItemFactory(SQLAlchemyModelFactory):
     summary = 'Summary of the feed item.'
     text = 'Text of the feed item.'
     feed = factory.SubFactory(FeedFactory)
+    totalVotes = 0
```
alexgerstein: Don't need this since we'll set default to 0.

sourcemash/users/models.py
```diff
@@ -8,7 +8,14 @@
...
 role_users = db.Table('roles_users',
         db.Column('user_id', db.Integer, db.ForeignKey('user.id')),
-        db.Column('role_id', db.Integer, db.ForeignKey('role.id')))
+        db.Column('role_id', db.Integer, db.ForeignKey('role.id'))
+)
+
+items_users = db.Table('items_users',
+        db.Column('user_id', db.Integer, db.ForeignKey('user.id')),
+        db.Column('item_id', db.Integer, db.ForeignKey('item.id')),
+        db.Column('vote', db.Integer)
```
alexgerstein: Make this default to 0.

sourcemash/users/models.py
```diff
@@ -24,6 +31,10 @@ class User(db.Model, UserMixin):
                             secondary=role_users,
                             backref=db.backref('users', lazy='dynamic'),
                             lazy='dynamic')
+    items = db.relationship('Item',
+                            secondary=items_users,
```
alexgerstein: Will need to change name of table here accordingly.

migrations/versions/426f1ad50ca2_.py
swglad: Followed the previous convention by which the role-user relationship is called role_user

alexgerstein: We call that table role_users. And it makes sense because there are more users than roles. It's a Western convention but I like the image of the shorter column being on the left side.

Similarly, we're likely to have less users than columns, so I think we should keep that convention.

Thoughts?

tests/factories/item_factories.py
swglad: Might be worth leaving in case we want to change it. Every other column of the model is represented here, it'd be weird to leave one out.

sourcemash/frontend/templates/category.tpl
```diff
@@ -19,6 +19,8 @@
           <span class="badge"><a href="#/categories/<%=item.get('category_2')%>"><%=item.get('category_2')%></a></span>
           <h4><a href="<%= '#/items/' + item.get('id') %>"><%= item.get('title') %></a></h4>
         </span>
+        <span class="waves-effect btn-floating red"><i class="mdi-action-thumb-down downvote"></i></span>
```
alexgerstein: We should be using class="row" and class"col m12" etc. to make it look nicer on the phone. Up to you if you want to do that here or just add it as another issue.

swglad: Niceness of the frontend is an issue for another day. We'll end up doing a small design overhaul sometime in the near future to get everything looking nice, but it's not necessary for now since it's highly subject to change.

sourcemash/items/models.py
alexgerstein: If you can upvote and downvote, totalVotes really represents the voteSum, not the voteCount. We should rename this accordingly. Also, do we want to maybe add a voteCount? It would be helpful to use votes as a way to pick the best article to insert, but we'll need a tally of votes to do this.

migrations/versions/40d101e1211c_.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '40d101e1211c'
+down_revision = '37e1e4549218'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
```
alexgerstein: Note how these migrations did nothing. You need to delete the migration files your commits added, rm your app.db and rerun upgrade then migrate.

sourcemash/frontend/templates/category.tpl
alexgerstein: Agreed.

tests/factories/item_factories.py
alexgerstein: That's what the default is for. If you want to build an item from the factory with a different number of votes, you'll be able to do ItemFactory(totalVotes=1000).

worker_tasks/scraper.py
```diff
@@ -52,7 +52,7 @@ def _store_items_and_category_counts(feed, categorizer):
 
         new_entry = Item(title=item.title, text=text, link=item.link, 
                             last_updated=item_last_updated, author=getattr(item, 'author', None),
-                            summary=getattr(item, 'summary', None), feed_id=feed.id)
+                            summary=getattr(item, 'summary', None), feed_id=feed.id, totalVotes=0)
```
alexgerstein: Don't need to declare total votes here either since we have the default.

tests/test_api/test_item.py
```diff
@@ -19,6 +26,38 @@ def test_get_item_missing(self, test_client, item):
         check_valid_header_type(r.headers)
         assert r.status_code == 404
 
+    def test_put_item_upvote(self, test_client, user, item):
+        self.login(test_client, user.email, user.password)
+
+        upvote = dict(vote=1)
```
alexgerstein: If we make vote a Bool, rather than an int (true=upvote, false=downvote), we won't have to worry about someone making a call to the api with >1 vote.

Or, maybe we can look into how to set a min and max on the vote?

Not sure what makes the most sense.

sourcemash/api/items.py
```diff
@@ -26,10 +26,23 @@ def get(self, feed_id):
...
     def get(self, feed_id, id):
         item = Item.query.get_or_404(id)
         return {'item': marshal(item, item_fields)}
 
+    @login_required
+    def put(self, feed_id, id):
+        ''' Update article vote count '''
+        args = self.reqparse.parse_args()
+        item = Item.query.get_or_404(id)
+        item.totalVotes += args.vote
```
alexgerstein: We should use the user_items db to check that a user has not voted too far in one direction (i.e. >1 or <-1).

migrations/versions/a5f4c2a45a3_.py
```diff
@@ -0,0 +1,34 @@
...
+revision = 'a5f4c2a45a3'
+down_revision = '598562686e8'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('users_items',
```
alexgerstein: user_items..

tests/test_api/test_item.py
swglad: Let's do integer and use your comment above with checking whether a user should be able to keep voting.

Also, disabling buttons on the frontend once you've voted is a good way to stop it from the user side.

sourcemash/api/items.py
alexgerstein: And of course write tests for this too!

tests/test_api/test_item.py
alexgerstein: Sounds good, we should still see if we can get the api to validate the integer is within a range. We don't want a malicious user trying to send a bigger number.

sourcemash/users/models.py
```diff
@@ -8,7 +8,14 @@
 
 role_users = db.Table('roles_users',
         db.Column('user_id', db.Integer, db.ForeignKey('user.id')),
-        db.Column('role_id', db.Integer, db.ForeignKey('role.id')))
+        db.Column('role_id', db.Integer, db.ForeignKey('role.id'))
+)
+
+users_items = db.Table('users_items',
```
swglad: should this be a db.Assocation?
http://docs.sqlalchemy.org/en/rel_0_9/orm/basic_relationships.html#many-to-many

sourcemash/api/items.py
swglad: I'm not sure how to do this since we need to get the third column from the db.Relationship....

alexgerstein: You should be able to get columns from that table using: User.items.has(vote=1) etc.

It's helpful to open python shell from manage.py and play with the possible queries to see what works.

swglad: I'll try doing that and replacing User with current_user

alexgerstein: I don't think you can do that. Try looking at my edited note above.

swglad:
***This PR is not complete until we update the scraper to add newly scraped items to each user to maintain the user_items relationship.***

alexgerstein:
Do we need a user->item relationship in the table for every item? Or just ones that are read/voted for? For example, to see if a user read an article, you can search by user_id and item_id and if there's no row in the table it means a) they didn't vote yet on the article and b) they didn't read the article yet. Then you can add the row here.

Thoughts?

swglad:
I thought we're adding an extra column to this table eventually so it's user_id-item_id-vote-unread. 

You should be able to read but not vote, so whether it's in the table can just tell you whether it's read or not. I'm not sure how you'd be able to see that a user read the article before pinging the table.

alexgerstein:
Yeah that's what I was thinking.

What I was trying to say (in response to your bolded comment) is we don't need to add every user-item relationship to the table during the scrape. We can just wait until it gets modified on the client side.

swglad:
That's fine. The issue for now is that we're going to start needing to use this table to get voting working, so until we have a client-side unread/read response, we need to add everything

alexgerstein:
I'm not sure I follow. We can't add the rows as needed during the PUT requests in the api?

manage.py
```diff
@@ -108,6 +108,12 @@ def seed():
     # Scrape articles for feed
     scrape_articles(Categorizer())
 
+    # Add items to user_item relationship
```
alexgerstein: We shouldn't have to add this here now since this was just for debugging.

sourcemash/api/items.py
```diff
@@ -17,18 +16,46 @@
     'category_1': fields.String,
     'category_2': fields.String,
     'summary': fields.String,
-    'category_1': fields.String,
-    'category_2': fields.String,
+    'totalVotes': fields.Integer,
```
alexgerstein: Let's rename this to score, or points, or something else that reflects that this is a fluctuating sum, not a count of number of votes.

sourcemash/api/items.py
```diff
@@ -17,18 +16,46 @@
...
         item = Item.query.get_or_404(id)
         return {'item': marshal(item, item_fields)}
 
+    @login_required
+    def put(self, id):
+        ''' Update item vote count '''
+        args = self.reqparse.parse_args()
+        item = Item.query.get_or_404(id)
+        
+        # Reject if size of vote is too large
```
alexgerstein: Extract these tests (abs(number) <= 1, requested vote != existing vote) to a form. Then call form validate on submit here. This will simplify the call here.

I extracted user_items out into a folder in sourcemash, so we can just add a forms.py to that folder now.

sourcemash/items/models.py
```diff
@@ -11,7 +11,7 @@ class Item(db.Model):
     feed_id = db.Column(db.Integer, db.ForeignKey('feed.id'))
     category_1 = db.Column(db.String(100))
     category_2 = db.Column(db.String(100))
-
+    totalVotes = db.Column(db.Integer, default=0)
```
alexgerstein: See above comment about making this name better.

tests/factories/item_factories.py
alexgerstein: We should remove this line again. I don't think we need it since the factory will use the default of 0 on the item model.

tests/fixtures/item_fixtures.py
```diff
@@ -15,4 +15,12 @@ def itemWithCategory(request):
 
 @pytest.fixture()
 def itemsWithCategory(request):
-	return [item_factories.ItemWithCategoryFactory() for i in range(5)]
\ No newline at end of file
+	return [item_factories.ItemWithCategoryFactory() for i in range(5)]
+
+@pytest.fixture()
```
alexgerstein: Don't need this fixture since it's covered by user_item_upvote now.

tests/fixtures/item_fixtures.py
```diff
@@ -15,4 +15,12 @@ def itemWithCategory(request):
...
+def itemWithUpvote(request):
+	return item_factories.ItemFactory(totalVotes=1)
+
+@pytest.fixture()
+def itemWithDownvote(request):
```
alexgerstein: Don't need this fixture since it's covered by user_item_downvote now.

sourcemash/api/items.py
```diff
@@ -17,18 +16,46 @@
...
-    'category_2': fields.String,
+    'totalVotes': fields.Integer,
     'uri': fields.Url('api.item')
 }
 
 
 class ItemAPI(Resource):
+    def __init__(self):
+        self.reqparse = reqparse.RequestParser()
+        self.reqparse.add_argument('vote', type = int, required = True)
```
alexgerstein: PUT on item will also need to handle "mark_as_read" eventually.

We will not want to pass both {read: True} and {vote: 0} every time we mark an article as read. Reqparse can take a default argument if it's not present, so that way we can remove required.

This way a mark_as_read and vote PUT won't have to be handled any different. We just need to specify on line 52 that args.vote != 0 (or check this on the form once we write that).

Does this make sense?

sourcemash/api/items.py
```diff
@@ -17,18 +16,46 @@
...
 
+    @login_required
+    def put(self, id):
+        ''' Update item vote count '''
+        args = self.reqparse.parse_args()
+        item = Item.query.get_or_404(id)
+        
+        # Reject if size of vote is too large
+        if abs(args.vote) > 1:
+            return {'item': marshal(item, item_fields)}, 422
```
alexgerstein: This should return a dictionary of errors like below, but that will be covered once we write a form to validate the request.

tests/factories/user_item_factories.py
```diff
@@ -0,0 +1,23 @@
...
+    FACTORY_FOR = UserItem
+    FACTORY_SESSION = db.session
+
+    user = factory.SubFactory(UserFactory)
+    item = factory.SubFactory(ItemFactory)
+
+
+class UserItemUpvoteFactory(UserItemFactory):
+    vote = 1
+    item = factory.SubFactory(ItemFactory, totalVotes=1)
```
swglad: item.totalVotes = 1

tests/factories/user_item_factories.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class UserItemDownvoteFactory(UserItemFactory):
+    vote = -1
+    item = factory.SubFactory(ItemFactory, totalVotes=-1)
```
swglad: item.totalVotes += -1

manage.py
swglad: I think we should do it for one of the feeds so we have it for debugging purposes. I agree that it doesn't need to be done for all of them.

tests/factories/user_item_factories.py
swglad: Doesn't work

tests/factories/user_item_factories.py
swglad: Doesn't work

manage.py
alexgerstein: What do we need to debug with it?

-- 
Alex Gerstein
gerstein.alex@gmail.com

On February 20, 2015 at 2:40:12 PM, Scott Gladstone (notifications@github.com) wrote:

I think we should do it for one of the feeds so we have it for debugging purposes. I agree that it doesn't need to be done for all of them.

tests/factories/user_item_factories.py
alexgerstein: Try something like this? http://factoryboy.readthedocs.org/en/latest/reference.html#postgenerationmethodcall

manage.py
swglad: Testing queries mostly. It doesn't cost much to fill up the table with one user and one set of feed items, so I see little reason not to do it.

sourcemash/api/items.py
```diff
@@ -18,16 +18,46 @@
...
+                                                item_id=id).one()
+        except:
+            user_item = UserItem(user_id=current_user.id, item_id=id, vote=0)
+
+        if user_item.vote == args.vote:
+            return {'errors': {'vote': ["You have already voted on this item."]}}, 422
+
+        user_item.vote += args.vote
+        item.totalVotes += args.vote
+        db.session.commit()
```
swglad: does this need a db.session.add(user_item) ?

manage.py
alexgerstein: Ok sounds good.

-- 
Alex Gerstein
gerstein.alex@gmail.com

On February 20, 2015 at 2:57:39 PM, Scott Gladstone (notifications@github.com) wrote:

In manage.py:

> @@ -108,6 +108,12 @@ def seed():
>      # Scrape articles for feed
>      scrape_articles(Categorizer())
>   
> +    # Add items to user_item relationship
Testing queries mostly. It doesn't cost much to fill up the table with one user and one set of feed items, so I see little reason not to do it.

sourcemash/api/items.py
alexgerstein: Yeah. We can put that inside the except block where we create a new UserItem.

sourcemash/api/items.py
swglad: That'll mean we don't need reqparse at all then, correct?

alexgerstein: You'll still need reqparse. See how we use the feed form to create new subscriptions.

sourcemash/api/items.py
```diff
@@ -18,16 +19,50 @@
...
+        # Check vote column of user_items table
+        try:
+            user_item = UserItem.query.filter_by(user_id=current_user.id, 
+                                                item_id=id).one()
+        except:
+            user_item = UserItem(user_id=current_user.id, item_id=id, vote=0)
+            db.session.add(user_item)
+
+        if args.vote != 0: # not voting if vote = 0: unread/read check
+            if user_item.vote == args.vote:
```
alexgerstein: ```if args.vote and user_item.vote == args.vote```

don't need line 57.

sourcemash/api/items.py
```diff
@@ -18,16 +19,50 @@
...
+        # Reject if user has already voted
+        # Check vote column of user_items table
+        try:
+            user_item = UserItem.query.filter_by(user_id=current_user.id, 
+                                                item_id=id).one()
+        except:
+            user_item = UserItem(user_id=current_user.id, item_id=id, vote=0)
+            db.session.add(user_item)
+
+        if args.vote != 0: # not voting if vote = 0: unread/read check
```
alexgerstein: Don't need ```!= 0```

sourcemash/api/items.py
```diff
@@ -18,16 +19,50 @@
...
+        item = Item.query.get_or_404(id)
+        
+        # Reject if user has already voted
+        # Check vote column of user_items table
+        try:
+            user_item = UserItem.query.filter_by(user_id=current_user.id, 
+                                                item_id=id).one()
+        except:
+            user_item = UserItem(user_id=current_user.id, item_id=id, vote=0)
+            db.session.add(user_item)
```
alexgerstein: Need to commit to DB here to set the vote default vote on the UserItem.

manage.py
```diff
@@ -108,6 +108,11 @@ def seed():
     # Scrape articles for feed
     scrape_articles(Categorizer())
 
+    # Add some items to user_item relationship
+    for item in techcrunch.items:
+        user.items.append(item)
```
alexgerstein: Will this work?

sourcemash/api/items.py
```diff
@@ -18,16 +19,51 @@
...
+
+        if not form.validate():
+            return {"errors": form.errors}, 422
+
+        item = Item.query.get_or_404(id)
+        
+        # Reject if user has already voted
+        # Check vote column of user_items table
+        try:
+            user_item = UserItem.query.filter_by(user_id=current_user.id,
```
alexgerstein: Filter by user and item here too.

sourcemash/user_items/forms.py
```diff
@@ -0,0 +1,12 @@
+from wtforms import IntegerField
+from wtforms.validators import ValidationError
+
+from flask_wtf import Form
+
+class VoteForm(Form):
+    vote = IntegerField('vote')
+
+    def validate_vote(form, field):
```
alexgerstein: We don't need to write a validate vote here. Instead, we can use the ```NumberRange(min=None, max=None, message=None)``` validator that comes from wtforms.

Issue #74, opened 2015-02-10
Introduce unsubscribed content


swglad:
Show articles to a user that are from a feed that they are not subscribed to.

swglad:
Select a random item out of the same CATEGORY such that user is not subscribed to that feed.

alexgerstein:
All articles in a category are already being returned. We need to create a new category table (see #110) and filter on feed id in user_feed table.

PR #119, opened 2015-02-19
Fix API to start working on subliminal articles.

alexgerstein:
This paves the way to work on #74.


sourcemash/api/categories.py
```diff
@@ -1,35 +1,55 @@
...
     def get(self):
-        categories = Counter(dict(Item.query.with_entities(Item.category_1, func.count()).group_by(Item.category_1).all()))
-        categories.update(dict(Item.query.with_entities(Item.category_2, func.count()).group_by(Item.category_2).all()))
+        user_feed_ids = [feed.id for feed in current_user.subscribed]
+
+        categories = Counter()
+
+        distinct_category_1 = Item.query.with_entities(Item.category_1, func.count())   \
+                                        .filter(Item.feed_id.in_(user_feed_ids))        \
+                                        .group_by(Item.category_1)                      \
```
swglad: Remind me why both with_entites and group_by are necessary again?

alexgerstein: With_Entities replaces SELECT * with SELECT cat_1, count

swglad: And you need group_by on top of that because..

alexgerstein: If you don't do group by, you'll just get the {Category_1}, Count for each item in the list and categories will repeat. You need to do group_by to make sure each category only appears once and its count is the total of all appearances of the category.

sourcemash/api/items.py
```diff
@@ -20,16 +22,44 @@
...
+
+    def get(self, category):
+        category = category.title()
+        items = Item.query.filter((Item.category_1 == category) | (Item.category_2 == category)).all()
+        return {'items': [marshal(item, item_fields) for item in items]}
+
+
+api.add_resource(ItemAPI, '/items/<int:id>', endpoint='item')
+api.add_resource(FeedItemListAPI, '/feeds/<int:feed_id>/items', endpoint='feed_items')
+api.add_resource(CategoryItemListAPI, '/categories/<string:category>/items', endpoint='category_items')
```
swglad: Why this and not:
/items/feeds/--id--
/items/categories/--category--
/items/categories/--category--/all

sourcemash/frontend/static/js/collections/items.js
```diff
@@ -1,12 +1,17 @@
 Sourcemash.Collections.Items = Backbone.Collection.extend({
     model: Sourcemash.Models.Item,
     url: function() {
-        return '/api/feeds/' + this.feed.get('id') + '/items'
+        if (this.feed) {
+            return '/api/feeds/' + this.feed.get('id') + '/items'
+        } else if (this.category) {
+            return '/api/categories/' + this.category.get('category') + '/items'
+        }
```
swglad: Is it every possible for both this.feed and this.category to be set simultaneously? How does it know which one to look at?

tests/fixtures/feed_fixtures.py
```diff
@@ -7,4 +7,8 @@ def feed(request):
 
 @pytest.fixture()
 def real_feed(request):
-    return feed_factories.TechCrunchFeedFactory()
\ No newline at end of file
+    return feed_factories.TechCrunchFeedFactory()
+
+@pytest.fixture()
+def feedWithItems(request):
+    return feed_factories.FeedFactory(items=[feed_factories.FeedFactory()])
```
swglad: How does this call actually give items to a feed factory? I see the function in FeedFactory(), but I'm unsure as to where items are actually being made/added.

tests/test_api/test_categories.py
```diff
@@ -1,45 +1,60 @@
...
+        assert set([data['categories'][0]['category'], data['categories'][1]['category']]) == \
+                set([itemWithCategory.category_1, itemWithCategory.category_2])
 
-class TestCategoryItemListAPI:
 
-    def test_get_category_items_present(self, test_client, itemWithCategory):
-        r = test_client.get('/api/categories/' + itemWithCategory.category_1)
+class TestCategoryListAPI(TestBase):
+    def test_get_user_categories_unsubscribed(self, test_client, user):
+        self.login(test_client, user.email, user.password)
```
swglad: something like TestBase.login isn't required to use this? How does self.login even work here?

sourcemash/api/feeds.py
```diff
@@ -18,14 +20,75 @@
 
 class FeedListAPI(Resource):
```
swglad: Should there be a delete to actually remove a feed from the database? It's not something that user's should be able to do, but maybe admins? For example, if an admin wanted to remove an "offensive" feed?

tests/test_api/test_categories.py
swglad: Whoops, sorry read the old class definition and missed the new one where you passed it. Got it!

sourcemash/api/feeds.py
alexgerstein: That's a good point. We might want that as an option in the future. Unsubscribe is the more common call though, so I think it makes sense to leave it as the DELETE to FeedList for now.

If we want to add an admin delete in the future, we can add a feeds/id/delete URL with role protection (admins-only) on it down the line.

sourcemash/api/items.py
alexgerstein: The standard paths for RESTFUL is to have an id immediately follow a resource if you want one model in particular, and the resource you're handling (i.e. items) should be the last one listed so it's clearer. For the first point, an outside would be confused how we get categories from items without defining which item to look at.

sourcemash/api/feeds.py
swglad: Unsubscribe is currently a delete in FeedAPI. We'd put admin delete in FeedListAPI, and perhaps have the admin delete button send the name of the feed in a form?

swglad: Would you want to add it now? I'd only encourage it because we're making enormous API changes here so it makes sense as part of this PR more than as a future issue.

sourcemash/frontend/static/js/collections/items.js
alexgerstein: We'd have to pass {feed: feed, category: category} for this to happen.

I'm trying to think of a time when a collection of items is filtered by both a feed and a category. What I came up with is this: if we want to allow users to filter a feed's items by category (or filter a category's items by feed), it would probably make more sense to have this dealt with as a filter function on the collection (see: http://stackoverflow.com/questions/11762105/filter-backbone-collection-by-attribute-value).

So I don't think we will have both this.feed and this.category. That being said, if both happened to be defined by some bug it would just resort to using the feed_id, which I think is satisfactory.

tests/fixtures/feed_fixtures.py
alexgerstein: That's a typo. Thanks for the catch.

sourcemash/api/feeds.py
alexgerstein: You don't want to put a DELETE on a list, since that route doesn't specify which feed you're trying to delete. Since we don't have any other admin commands yet--and deleting inappropriate feeds is more of an enhancement--I think we can take time to consider how we handle our admin API and create an "admin permissions" task for the future.

Also remember we don't even have ways of assigning people the admin role yet.

tests/test_api/test_items.py
```diff
@@ -0,0 +1,99 @@
...
+        data = json.loads(r.data)
+        assert len(data['items']) == 5
+
+
+    def test_get_users_items_case_insensitive_categories(self, test_client, userWithPopulatedFeed):
+        self.login(test_client, userWithPopulatedFeed.email, userWithPopulatedFeed.password)
+
+        feed = userWithPopulatedFeed.subscribed.first()
+
+        r = test_client.get('/api/categories/' + feed.items[0].category_1.lower() + '/items')
```
swglad: API calls are case insensitive? cool.

tests/test_api/test_items.py
```diff
@@ -0,0 +1,99 @@
...
+        assert r.status_code == 200
+
+        data = json.loads(r.data)
+        assert len(data['items']) == 5
+
+
+    def test_get_users_matching_categories_missing(self, test_client, userWithPopulatedFeed):
+        self.login(test_client, userWithPopulatedFeed.email, userWithPopulatedFeed.password)
+
+        outside_item = item_factories.ItemWithCategoryFactory(category_1="vegetables")
```
swglad: "vegetables" --> "nonexistent_category"

tests/test_api/test_items.py
```diff
@@ -0,0 +1,99 @@
...
+        r = test_client.get('/api/categories/' + feed.items[0].category_1.lower() + '/items')
+
+        check_valid_header_type(r.headers)
+        assert r.status_code == 200
+
+        data = json.loads(r.data)
+        assert len(data['items']) == 5
+
+
+    def test_get_users_matching_categories_missing(self, test_client, userWithPopulatedFeed):
```
swglad: Can you change the function name? The word 'matching' seems not right if you're just trying to get items for a nonexistent category.

PR #160, opened 2015-03-20
Category enhancements

### LINKED ISSUE #106: New ML approach to keyword generation.
Is there a better way to do this?
@alexgerstein @swglad

alexgerstein:
Improve categorization to rely less on heuristics. Instead, this uses Wikipedia to normalize categories and uses clusters to identify important keywords. Resolves #106.


worker_tasks/categorize.py
```diff
@@ -46,109 +60,368 @@
...
+        # Remove 's
+        word = word.replace("'s", "")
+
+        # Remove Smart Quotes
+        word = word.replace(u"\u2018", "").replace(u"\u2019", "").replace(u"\u201c","").replace(u"\u201d", "")
+
+        return word
+
+
+    def _scrape_wiki_links(self, titles):
```
alexgerstein: This is the messiest function. It's a scrape so it's hard to make it more readable. Let me know if you see anything...

alexgerstein:
Currently, this still takes only the best two categories. I figured I'd make the DB changes to assign more than 2 categories its own issue.

.coveragerc
```diff
@@ -1,8 +1,17 @@
 # .coveragerc to control coverage.py
 
+[run]
+omit =
+    config/*
+    flask/*
+    instance/*
+    migrations/*
+    sourcemash/frontend/*
+    tests/*
```
swglad: is coverage supposed to omit tests/?

.travis.yml
```diff
@@ -1,7 +1,12 @@
 language: python
+cache:
+  directories:
+    - $PWD/wheelhouse
```
swglad: wheelhouse??

config/default.py
```diff
@@ -25,6 +25,11 @@
 SECURITY_MSG_PASSWORD_NOT_PROVIDED = ("Invalid username/password combination", "error")
 SECURITY_MSG_USER_DOES_NOT_EXIST = ("Invalid username/password combination", "error")
 
-logging.basicConfig(level=logging.DEBUG)
+logging.basicConfig(level=logging.WARNING)
```
swglad: What do these logger settings do differently than before? Why WARNING instead of DEBUG?

.coveragerc
alexgerstein: Yeah. If you omit in coverage, it still runs the tests; it just doesn't return the % covered of the files in those folders.

We don't want to track 100% coverage on our test code...or on entire Python libraries for that matter...

manage.py
```diff
@@ -22,15 +22,10 @@
 app = create_app(os.environ.get("APP_CONFIG_FILE") or "development")
 manager = Manager(app)
 
-TEST_CMD = "py.test --cov-config .coveragerc --cov sourcemash \
+TEST_CMD = "py.test --cov-report term-missing --cov-config .coveragerc --cov . \
```
swglad: Why ```.``` instead of ```sourcemash```? What's ```--cov-report term-missing```?

config/default.py
alexgerstein: The level is the minimum level for the logger to print (to stdout). By making it WARNING instead of DEBUG, we're saying don't write it out to the console unless it's worth knowing. You'll see that for the development configuration, we lower it back down to DEBUG so we can see what's going on. When we look at Heroku logs though, we don't need to see everything, just the relevant stuff.

You can take a look at the levels here: https://docs.python.org/2/library/logging.html#levels.

Speaking of which, instead of writing print statements when debugging, we should write logger.debug("WHAT IS THIS VALUE???").

.travis.yml
alexgerstein: https://wheel.readthedocs.org/en/latest/

manage.py
```diff
@@ -51,19 +46,14 @@ def test(all=False):
...
         logging.info("Starting scrape...")
-        if (iterations % ITERATIONS_BEFORE_RESET) == 0:
-            categorizer.reset_title_categories()
 
         scrape_articles(categorizer)
 
-        iterations += 1
-        logging.info("Finished scrape. Zzz...")
-        time.sleep(SCRAPE_INTERVAL)
+        logging.info("Finished scrape. Let's run it back...")
```
swglad: Why did you get rid of sleep? Wouldn't this just be an infinite repeating loop that never stops?

tests/conftest.py
```diff
@@ -1,6 +1,8 @@
 from fixtures.app_fixtures import *
+from fixtures.selenium_fixtures import *
+from fixtures.categorizer_fixtures import *
+
```
swglad: Remove \n

.coveragerc
swglad: Definitely agree with libraries. Just checking in on tests.

.travis.yml
swglad: Cool

manage.py
alexgerstein: worker_tasks isn't in the sourcemash folder. So I broadened the scope of pytest, but specified what to omit in the coveragerc (as you saw above).

term-missing will print out both the percentages and the missing lines so you don't need to run coverage report each time after running tests.

config/default.py
swglad: That's what I figured but looks good.

manage.py
alexgerstein: Yeah. Isn't that what our scraper is supposed to do?

The new categorization runs slower, so sleeping will just add more time to an already slow process. Is there a reason we need a sleep at the moment since this is running on its own process?

manage.py
swglad: *thumbs up*

manage.py
swglad: I guess not. Mainly because it requires more cycles when it's running than when it's sleeping, but if you say so. It's only good not to sleep if the time to categorize new articles equals the time to scrape and categorize the previous set of new articles.

tests/test_worker/test_categorization.py
```diff
@@ -1,101 +1,47 @@
...
-        assert categorizer.get_valid_ngrams("Harry's") == ["Harry"]
-
-
-    def test_get_valid_ngrams_punctuation(self):
-        categorizer = Categorizer()
-        assert categorizer.get_valid_ngrams('Google.') == ["Google"]
+    def test_categorize_item(self, categorizer, ebolaItem):                
+        categories = categorizer.categorize_item(ebolaItem.title, ebolaItem.text)
+        overlapping_categories = filter(lambda x: "Ebola" in x or "West Africa" in x, categories)
+        assert len(overlapping_categories) == 2
```
swglad: Clever written test. Nice.

manage.py
alexgerstein: I'm not sure I follow this comment. Wouldn't this never be true?

swglad: Effectively you're scraping the web for new articles as soon as you finish categorizing the most recent ones. Is this efficient? Or necessary?

alexgerstein: It's not necessary, per sé. We could put the sleep back, but I don't think that would make the infinite loop any more efficient, do you?

tests/test_worker/test_categorization.py
```diff
@@ -1,101 +1,47 @@
...
+        assert categorizer._is_viable_candidate("100") == False
 
+    def test_is_viable_candidate_tuple_both_valid(self, categorizer):
+        assert categorizer._is_viable_candidate("elemental alligator") == True
 
-    def test_is_valid_category_tuple_both_valid(self):
-        categorizer = Categorizer()
-        assert categorizer.is_valid_category(('elemental alligator')) == True
+    def test_is_viable_candidate_tuple_one_invalid(self, categorizer):
+        assert categorizer._is_viable_candidate("and alligator") == False
```
swglad: How does something like "Abercrombie and Fitch" work? Can you add a couple 3-gram tests?

tests/test_worker/test_categorization.py
```diff
@@ -1,101 +1,47 @@
...
 
+    def test_decode_list_unicode(self, categorizer):
+        assert categorizer._decode_list([u"éxit"]) == ["éxit"]
 
-    def test_is_valid_category_tuple_one_invalid(self):
-        categorizer = Categorizer()
-        assert categorizer.is_valid_category(('and alligator')) == False
\ No newline at end of file
+    def test_decode_list_list(self, categorizer):
+        assert categorizer._decode_list([["String"]]) == [["String"]]
```
swglad: Can you add multiple items if it's decoding a list?

tests/test_worker/test_categorization.py
alexgerstein: I was hoping to keep the number of tests that require Wikipedia down to a minimum since they're slower, but I can add it if you'd like.

Let me know.

worker_tasks/scraper.py
```diff
@@ -30,9 +32,15 @@ def scrape_articles(categorizer):
 
         # Extract text and categorize item
         text_only = soup.get_text()
-        item.category_1, item.category_2 = categorizer.categorize_item(item.title, text_only)
+        categories = categorizer.categorize_item(item.title, text_only)
+        if len(categories) >= 1:
+            item.category_1 = categories[0]
+
+        if len(categories) >= 2:
+            item.category_2 = categories[1]
```
swglad: So when we have a category table going, this will just be iterating over ``categories`` and adding CategoryItems to the category_item table?

alexgerstein: Correct.

worker_tasks/scraper.py
```diff
@@ -42,17 +50,12 @@ def _get_full_text(url):
 
 
 def _store_items_and_category_counts(feed, categorizer):
```
swglad: Rename ```_store_items```

worker_tasks/scraper.py
```diff
@@ -11,6 +11,8 @@
 
 import logging
 
+logger = logging.getLogger('Sourcemash')
+
 def scrape_articles(categorizer):
     # Pull down all articles from RSS feeds
     for feed in Feed.query.all():
```
swglad: Rename function call to ```_store_items``` since it doesn't store category counts anymore.

worker_tasks/scraper.py
swglad: Also doesn't need a ```categorizer``` anymore.

manage.py
swglad: We don't exactly need "sleep" but more like "pass control on to other functions for a while and don't waste your computation energy on the scraper for now". Aka redis queue...

swglad: This is tabled for now and re-adding a queue is an issue for another day.

tests/test_worker/test_categorization.py
alexgerstein: Sure.

tests/test_worker/test_categorization.py
alexgerstein: Sure.

tests/test_worker/test_categorization.py
swglad: Definitely add one for a trigram that we'd want to keep (e.g. Crate & Barrel, Los Pollos Hermanos). One is sufficient.

tests/test_worker/test_categorization.py
```diff
@@ -1,101 +1,53 @@
...
+        assert categorizer._is_viable_candidate("Ambercrombie and Fitch") == True
 
-    def test_is_valid_category_tuple_both_valid(self):
-        categorizer = Categorizer()
-        assert categorizer.is_valid_category(('elemental alligator')) == True
+    def test_decode_list_unicode(self, categorizer):
+        assert categorizer._decode_list([u"éxit"]) == ["éxit"]
 
+    def test_decode_list_list(self, categorizer):
+        assert categorizer._decode_list([["String"]]) == [["String"]]
```
swglad: This necessary if you have the test below it?

tests/test_worker/test_categorization.py
```diff
@@ -1,101 +1,53 @@
...
+        assert categorizer._is_viable_candidate("elemental alligator") == True
 
-    def test_is_valid_category_number(self):
-        categorizer = Categorizer()
-        assert categorizer.is_valid_category('100') == False
+    def test_is_viable_candidate_bigram_one_invalid(self, categorizer):
+        assert categorizer._is_viable_candidate("and alligator") == False
 
+    def test_is_viable_candidate_trigram_valid(self, categorizer):
+        assert categorizer._is_viable_candidate("Ambercrombie and Fitch") == True
```
swglad: *Abercrombie

worker_tasks/categorize.py
```diff
@@ -1,11 +1,44 @@
...
+# 
+# Adapted from algorithm by Grineva, et. al. presented at:
+# http://www.slideshare.net/maria.grineva/extracting-key-terms-from-noisy-and-multitheme-documents?related=1
+# 
+# Algorithm Overview
+# 1) Get initial list of possible keyword candidates using ngram frequencies
+# 2) For each keyword candidate, use Wikipedia to extract all related article titles
+#       a) Related articles are called (disambiguation links) on Wikipedia
+# 3) Try to assign one Wikipedia article to each keyword candidate
+#       a) First, assign articles to the keywords with only 1 related article
```
swglad: with --> that have. Also, can you add an example?

worker_tasks/categorize.py
```diff
@@ -1,11 +1,44 @@
...
+#       b) Use these "auto-assigned" articles as the context to narrow the other categories' article
```
swglad: "context" meaning the universe of possible categories? Or can we still get categories that are outside the set of all "auto assigned" article titles?

worker_tasks/categorize.py
```diff
@@ -1,11 +1,44 @@
...
+#       c) With the context from (b), compute relatedness scores to find the best article match
+#           i) relatedness score = [SUM OF SHARED WIKIPEDIA HYPERLINKS] / [TOTAL NUMBER OF WIKIPEDIA HYPERLINKS]
+# 4) With each keyword assigned 1 article, create a graph where each vertex is an article and the edge weight
```
swglad: Can a keyword be assigned more than 1 article by this step? Or can it be assigned zero?

worker_tasks/categorize.py
alexgerstein: The context is the set of all "auto-assigned" titles. We can't go outside this set because we need  actual articles to compute relevance, since that score relies on the number of overlapping Wikipedia articles.

worker_tasks/categorize.py
```diff
@@ -1,11 +1,44 @@
...
+#       is the relatedness score between two articles
+# 5) Analyze the graph from (4) to identify keyword clusters
+# 6) Only keep the clusters that have a sufficient number of keywords that contain an original keyword candidate from (1)
+# 8) Return the best keywords
+#       a) For now, we still pick the top two keywords by frequency from kept clusters, but...
+#       b) TODO: WE WILL KEEP ALL KEYWORDS FROM THE KEPT CLUSTERS
```
swglad: Awesome documentation.

worker_tasks/categorize.py
```diff
@@ -1,11 +1,44 @@
...
+from bs4 import BeautifulSoup
 
-BIGRAM_WEIGHT = 1.25
-CAPITALIZED_WORD_WEIGHT = 1.5
+import igraph
+
+logger = logging.getLogger('Sourcemash')
+
+NGRAMS = 3
+MINIMUM_RELATEDNESS_SCORE = 0.1
```
swglad: What kind of scale do relatedness scores work on? Can you give a range of possible scores (e.g. [0,1], [-inf, inf], [0,10+] meaning that anything past 10 is enormous, etc.)?

worker_tasks/categorize.py
```diff
@@ -46,109 +79,374 @@
...
     def __init__(self):
-        self.title_categories = Counter()
+        self._memoized_related_articles = defaultdict(list)
+        self._memoized_article_links = defaultdict(list)
+        self._memoized_semantic_relatedness_scores = defaultdict(float)
+
 
     def categorize_item(self, title, text):
-        categories = Counter()
+        # Use old word count approach to get possible keywords
```
swglad: old --> ngram

worker_tasks/categorize.py
```diff
@@ -46,109 +79,374 @@
...
+        for ngram, count in ngrams.iteritems():
+            
+            # ...from title (x2)
+            if ngram in title_ngrams:
+                ngrams.update({ngram: count})
 
-        one_day_ago = datetime.now() - timedelta(hours=24)
-        titles = [item.title for item in Item.query.filter(Item.category_1==None, Item.last_updated>=one_day_ago).all()]
-        self.parse_title_categories(titles)
+            # ...bigram (x2) or trigram (x4)
```
swglad: ...^from

swglad: Why x4 for trigram? Just curious since we're weighting trigrams twice as much as titles and that seems like a lot offhand. But you're the one that wrote this so it's totally your judgment call.

worker_tasks/categorize.py
```diff
@@ -46,109 +79,374 @@
...
+            for n in range(ngram.count(' ')):
```
swglad: Is something like ```range(len(ngram))``` possible?

worker_tasks/categorize.py
alexgerstein: It's x3. I wrote it wrong in the comment.

worker_tasks/categorize.py
```diff
@@ -46,109 +80,374 @@
...
+                ngrams.update({ngram: count})
```
swglad: Shouldn't this be using ```n: count``` to update?

worker_tasks/categorize.py
swglad: I'm slightly confused how this for loop works...

worker_tasks/categorize.py
```diff
@@ -46,109 +80,374 @@
...
 
+        return dict(ngrams.most_common(20))
```
swglad: Define 20 above as a "magic number"?

worker_tasks/categorize.py
alexgerstein: That wouldn't work since ngram is just a string. Length of a string would be the number of characters. We could do a split to break the ngram into an array of the grams, but that seemed unnecessary.

worker_tasks/categorize.py
```diff
@@ -46,109 +80,374 @@
...
+            
+            # ...from title (x2)
+            if ngram in title_ngrams:
+                ngrams.update({ngram: count})
 
-        one_day_ago = datetime.now() - timedelta(hours=24)
-        titles = [item.title for item in Item.query.filter(Item.category_1==None, Item.last_updated>=one_day_ago).all()]
-        self.parse_title_categories(titles)
+            # ...from bigram (x2) or trigram (x3)
+            for n in range(ngram.count(' ')):
```
alexgerstein: N is an integer in range from (0..{NUMBER OF SPACES IN NGRAM - 1}).

Count returns the number of appearances of a string in a word. When we call count(' ') on an ngram, we're getting the number of spaces in the ngram. The number of spaces in an ngram is equal to one less the number of words in the ngram.

For example, when we have a bigram (i.e. "Stop Words"), the loop [n] will happen once.

worker_tasks/categorize.py
```diff
@@ -46,109 +80,374 @@
...
+        for n in range(NGRAMS):
+            ngrams = zip(*[words[i:] for i in range(n + 1)])
+
+            for ngram in ngrams:
+                ngram = ' '.join(ngram)
+                if self._is_viable_candidate(ngram):
+                    valid_ngrams.update([ngram])
+
+        return valid_ngrams
+
```
swglad: Can you move ```_clean_word``` right below here since this is the only place you call it?

alexgerstein: Sure. Makes sense.

worker_tasks/categorize.py
```diff
@@ -46,109 +80,374 @@
...
+                item = item.encode('utf-8')
+            elif isinstance(item, list):
+                item = self._decode_list(item)
+            elif isinstance(item, dict):
+                item = self._decode_dict(item)
+            rv.append(item)
+        return rv
+
+
+    def _decode_dict(self, data):
```
swglad: This isn't explicitly tested? Should it be? I only ask because _decode_list is

worker_tasks/categorize.py
swglad: Yup, perfect. Figured it out after you made the initial change. Looks good.

worker_tasks/categorize.py
alexgerstein: Since these two functions are pretty trivial, I tried to cover them all in as few tests as possible. The list contains a dict, which is how I got the coverage to pass with 100%. Figured it wasn't worth writing too many tests over.

swglad: Sounds good. No more work necessary. Nothing lost, nothing gained.

worker_tasks/categorize.py
```diff
@@ -46,109 +81,374 @@
...
+
+            for ngram in ngrams:
+                ngram = ' '.join(ngram)
+                if self._is_viable_candidate(ngram):
+                    valid_ngrams.update([ngram])
+
+        return valid_ngrams
+
+
+    def _clean_word(self, word):
```
swglad: Does wikipedia deal with capitalization? No need for .lower()?

worker_tasks/categorize.py
```diff
@@ -46,109 +81,374 @@
...
-            if self.is_bigram(category):
-                categories[category] *= BIGRAM_WEIGHT
+        logger.debug("\nMemoizing Articles Links...")
 
-            if category.istitle():
-                categories[category] *= CAPITALIZED_WORD_WEIGHT
+        article_titles = [title for ngram in ngrams for title in self._memoized_related_articles[ngram]]
+        self._scrape_wiki_links(article_titles)
 
-        return categories
```
swglad: Move ```_scrape_wiki_links``` here

worker_tasks/categorize.py
```diff
@@ -46,109 +81,374 @@
...
+            if titled_words not in [[True, False, True], [True, True,True]]:
+                return False 
+
+        return True
+
+
+    def _scrape_wiki_links(self, titles):
+        unscraped_titles = filter(lambda x: x not in self._memoized_article_links, titles)
+
+        for i in xrange(0, len(unscraped_titles), 5):
```
swglad: 5 determined heuristically or did wiki have recommendations on batch sizes?

worker_tasks/categorize.py
alexgerstein: Yeah. That's what the Normalization and Redirects are.

worker_tasks/categorize.py
alexgerstein: Wiki didn't have any recommendations; I just came up with 5. I think 10 was too much and I didn't want to push it for fear that we crash the scraper.

worker_tasks/categorize.py
```diff
@@ -46,109 +80,349 @@
...
+
+        shared_links_count = 0
+        for article_link in article_1_links:
+            weight = 1
+            if "Template" in article_link or \
+                "Wikipedia" in article_link or \
+                "Help" in article_link:
+                weight = 0.1
+            if "Category" in article_link:
+                weight = 1.5
```
swglad: Can you add comments to explain why these should be weighted as such?

alexgerstein: It's not explained by the terms (i.e. "Help" and "Template")?

worker_tasks/categorize.py
```diff
@@ -46,109 +80,349 @@
...
+                    continue
 
-    def is_bigram(self, word):
-        return " " in word
+                if (article_1, article_2) in self._memoized_semantic_relatedness_scores:
+                    relatedness_score = self._memoized_semantic_relatedness_scores[(article_1, article_2)]
+                else:
+                    relatedness_score = self._get_relatedness_score(self._memoized_article_links[article_1], self._memoized_article_links[article_2])
+                    self._memoized_semantic_relatedness_scores[(article_1, article_2)] = relatedness_score
+                    self._memoized_semantic_relatedness_scores[(article_2, article_1)] = relatedness_score
```
swglad: Can you index into python dictionaries using sets or is that not a thing? Long shot but saves a line if sets are hashable

alexgerstein: Sets are mutable in Python, so they can't be the key.

worker_tasks/categorize.py
```diff
@@ -46,109 +80,349 @@
...
 
+                if relatedness_score:
+                    edges.append([i, j])
+                    edge_attrs["weight"].append(relatedness_score)
 
-    def is_valid_category(self, category):
+        return igraph.Graph(edges=edges, vertex_attrs=vertex_attrs, edge_attrs=edge_attrs)
```
swglad: I assume this returns an undirected graph since you only append ```edge(i,j)``` and not ```edge(j,i)```? Is undirected the default of ```igraph.Graph``` constructor?

alexgerstein: It's directed, i and j will eventually swap since they both loop through the full articles array. The only one it skips is i == j (where the edge goes to itself).

swglad: Ahh makes sense. Thanks.

worker_tasks/categorize.py
swglad: Makes sense.

worker_tasks/categorize.py
swglad: Don't those mean the link is garbage and should effectively add no weight? That's what I didn't really get. And what is "Category"?

worker_tasks/categorize.py
```diff
@@ -46,109 +80,349 @@
...
+
+        best_keywords = set()
+        keyword_counts = Counter(original_keywords)
+
+        for cluster in communities.subgraphs():
+            vertex_names = map(lambda x: x['name'], cluster.vs)
+
+            matching_vertices_count = sum(map(lambda x: keyword_counts[x] if x in keyword_counts else 0, vertex_names))
+
+            cluster_score = float(matching_vertices_count) / len(cluster.vs)
```
swglad: Won't this almost always be less than 1?

The only way it can't be is if there are words with keyword_counts[word] that are pretty big. Does that happen frequently?

worker_tasks/categorize.py
```diff
@@ -46,109 +80,349 @@
...
+
+            if cluster_score > 1:
+                best_keywords.update(vertex_names)
```
swglad: Do we want to just update best_keywords or do we want to see if  current_cluster_score is better than best_cluster_score and overwrite best_keywords if the current_cluster_score > best_cluster_score? (same as how you did max relatedness score - line 277)

worker_tasks/categorize.py
```diff
@@ -46,109 +80,349 @@
...
+        keyword_counts = Counter(original_keywords)
+
+        for cluster in communities.subgraphs():
+            vertex_names = map(lambda x: x['name'], cluster.vs)
+
+            matching_vertices_count = sum(map(lambda x: keyword_counts[x] if x in keyword_counts else 0, vertex_names))
+
+            cluster_score = float(matching_vertices_count) / len(cluster.vs)
+
+            if cluster_score > 1:
```
swglad: Why 1?

worker_tasks/categorize.py
swglad: Seems reasonable.

worker_tasks/categorize.py
swglad: There's actually a type called an ImmutableSet that can be used as a dictionary key, but I think it's unnecessary. 

In case you're interested: https://docs.python.org/2/library/sets.html

worker_tasks/scraper.py
```diff
@@ -11,17 +13,21 @@
 
 import logging
 
-def scrape_articles(categorizer):
+logger = logging.getLogger('Sourcemash')
+
+def scrape_and_categorize_articles():
+    categorizer = Categorizer()
```
swglad: To confirm, we can/are doing this because we're now using Wikipedia as the universe of possible categories and thus not maintaining an application-level notion of categories like we were before these changes?

worker_tasks/categorize.py
alexgerstein: It felt right! We can adjust when we start adding all categories in a cluster.

worker_tasks/scraper.py
alexgerstein: Correct-a-mundo.

worker_tasks/categorize.py
alexgerstein: There could be solid keywords in multiple clusters (I think Ebola/West Africa is one such example), so I'm not picking only the best cluster. Also, remember the Apple/Blindness example.

worker_tasks/categorize.py
alexgerstein: Keyword counts can get to be pretty high because the weights (for bigram, trigram, title, etc.) are still applied.

worker_tasks/categorize.py
alexgerstein: That's helpful to know. Thanks for the link.

worker_tasks/categorize.py
alexgerstein: This is an example of Category: http://en.wikipedia.org/wiki/Category:Comics

alexgerstein: That's a good point about the Template/Wikipedia links. We may want to just make the weight 0 on those.

worker_tasks/categorize.py
swglad: Cool

worker_tasks/categorize.py
swglad: Sounds good.

worker_tasks/categorize.py
swglad: *PLEASE* make Apple/Blindness a comment so I don't ask you this again (which I know I will do lol)!!! Say something like "update multiple clusters to accomodate multi-category articles such as "Apple Inc. solves Blindness with new iEye"

PR #163, opened 2015-03-20
Marshal fields

### LINKED ISSUE #151: Add detailed fields to marshalling for speed improvement
For example:
```
feed_fields = {
   title: fields.String
}
feed_fields_detailed = {
   fields.Nested(feed_fields),
   unread_count: getUnreadCount
}
```

(and of course change that in the api calls below to actually get the enhancement!)

swglad:
Resolves #151.

### DIFF
--- a/sourcemash/api/feeds.py
+++ b/sourcemash/api/feeds.py
@@ -39,10 +39,14 @@ feed_fields = {
     'description': fields.String,
     'image_url': fields.String,
     'last_updated': fields.DateTime,
-    'item_count': getItemCount,
-    'unread_count': getUnreadCount,
     'subscribed': isSubscribed
 }
+
+feed_fields_detailed = {
+    'item_count': getItemCount,
+    'unread_count': getUnreadCount
+}
+feed_fields_detailed.update(feed_fields)
 
 class FeedListAPI(Resource):
 
@@ -54,7 +58,7 @@ class FeedListAPI(Resource):
 
     @login_required
     def get(self):
-        return {'feeds': [marshal(feed, feed_fields) for feed in current_user.subscribed]}
+        return {'feeds': [marshal(feed, feed_fields_detailed) for feed in current_user.subscribed]}
 
 
     @login_required
@@ -87,7 +91,7 @@ class FeedListAPI(Resource):
             current_user.subscribed.append(feed)
             db.session.commit()
 
-        return marshal(feed, feed_fields), 201
+        return marshal(feed, feed_fields_detailed), 201
 
 
 class FeedListAllAPI(Resource):
--- a/sourcemash/api/feeds.py
+++ b/sourcemash/api/feeds.py
@@ -36,17 +36,21 @@ feed_fields = {
     'id': fields.Integer,
     'title': fields.String,
     'url': fields.String,
-    'description': fields.String,
-    'image_url': fields.String,
-    'last_updated': fields.DateTime,
     'subscribed': isSubscribed
 }
 
 feed_fields_detailed = {
+    'description': fields.String,
+    'image_url': fields.String,
+    'last_updated': fields.DateTime
+}
+feed_fields_detailed.update(feed_fields)
+
+feed_fields_status = {
     'item_count': getItemCount,
     'unread_count': getUnreadCount
 }
-feed_fields_detailed.update(feed_fields)
+feed_fields_status.update(feed_fields)
 
 class FeedListAPI(Resource):
 
@@ -58,7 +62,7 @@ class FeedListAPI(Resource):
 
     @login_required
     def get(self):
-        return {'feeds': [marshal(feed, feed_fields_detailed) for feed in current_user.subscribed]}
+        return {'feeds': [marshal(feed, feed_fields_status) for feed in current_user.subscribed]}
 
 
     @login_required
@@ -91,13 +95,13 @@ class FeedListAPI(Resource):
             current_user.subscribed.append(feed)
             db.session.commit()
 
-        return marshal(feed, feed_fields_detailed), 201
+        return marshal(feed, feed_fields_status), 201
 
 
 class FeedListAllAPI(Resource):
 
     def get(self):
-        return {'feeds': [marshal(feed, feed_fields) for feed in Feed.query.all()]}
+        return {'feeds': [marshal(feed, feed_fields_detailed) for feed in Feed.query.all()]}
 
 
 class FeedAPI(Resource):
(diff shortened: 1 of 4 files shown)


tests/test_api/test_feeds.py
```diff
@@ -15,16 +15,6 @@ def test_get_feed_present(self, test_client, feed):
 
         data = json.loads(r.data)
         assert data['feed']['title'] == feed.title
-        assert data['feed']['item_count'] == 0
-
-    def test_get_populated_feed_present(self, test_client, feedWithItems):
```
swglad: Really was just testing the fixture, which we know works. Now that the ```api/feeds/--id--``` call doesn't contain item_count, this test is redundant.

sourcemash/api/items.py
```diff
@@ -32,23 +34,20 @@ def output(self, key, item):
...
-    'summary': fields.String,
+    'voteSum': fields.Integer,
     'image_url': fields.String,
+    'summary': fields.String,
     'feed': fields.Nested(feed_fields),
     'unread': getUnreadStatus,
     'saved': getSavedStatus,
-    'vote': getVote,
-    'voteSum': fields.Integer,
-    'uri': fields.Url('api.item')
```
swglad: Deleted ```text``` and ```uri```. This file has no other changes.

sourcemash/api/items.py
```diff
@@ -10,6 +10,8 @@
 
 class getVote(fields.Raw):
     def output(self, key, item):
+        if not current_user.is_authenticated():
+            return 0
```
swglad: Items need to carry all of the information in item_fields in order to be rendered correctly. The only way to speed item rendering is to eliminate database calls/tries when a user isn't logged in. So, I added that for getVote and it is already present in getUnreadStatus and getSavedStatus.

swglad:
We get speed benefits for feed_fields because we render fields differently under different settings, but not really for item_fields because items have a common look/expectation across all pages (for now). See comments in PR.

sourcemash/api/feeds.py
```diff
@@ -24,25 +24,28 @@ def output(self, key, feed):
 class getUnreadCount(fields.Raw):
     def output(self, key, feed):
         total_item_count = Item.query.filter_by(feed_id=feed.id).count() 
-        
-        if not current_user.is_authenticated():
-            return total_item_count 
-
```
swglad: Not needed anymore because feed_fields_status only pings for this value on login_required requests.

sourcemash/api/feeds.py
```diff
@@ -24,25 +24,28 @@ def output(self, key, feed):
...
-        
         read_item_count = UserItem.query.filter_by(user=current_user, feed_id=feed.id, unread=False).count()
-
         return total_item_count - read_item_count
 
 feed_fields = {
     'id': fields.Integer,
     'title': fields.String,
     'url': fields.String,
+    'subscribed': isSubscribed
```
swglad: Always need ```subscribed``` for prompting user to subscribe & sign up if they're not logged in (I don't believe this behavior is currently covered - new issue?), and for user to subscribe on an unsubscribed feed if they are logged in.

swglad:
Coverage is 100%. Made a commit and fixed it.

sourcemash/api/feeds.py
alexgerstein: Sweet.

sourcemash/api/feeds.py
```diff
@@ -24,25 +24,28 @@ def output(self, key, feed):
...
+}
+
+feed_fields_detailed = {
     'description': fields.String,
     'image_url': fields.String,
-    'last_updated': fields.DateTime,
+    'last_updated': fields.DateTime
+}
```
alexgerstein: The fields in feed_fields_detailed won't require any more time to compile than the fields in feed_fields (i.e. they're just String and DateTime fields). So I think, to simplify things, we should keep description, image_url, and last_updated in the regular feed fields, and remove feed_fields_detailed.

sourcemash/api/feeds.py
```diff
@@ -24,25 +24,28 @@ def output(self, key, feed):
...
+feed_fields_detailed.update(feed_fields)
+
+feed_fields_status = {
     'item_count': getItemCount,
-    'unread_count': getUnreadCount,
-    'subscribed': isSubscribed
+    'unread_count': getUnreadCount
 }
+feed_fields_status.update(feed_fields)
```
alexgerstein: Since we "extend" from feed_fields, could we make this more readable by having feed_fields updated with feed_status_fields?

i.e.
```python
feed_status_fields = dict(feed_fields, **feed_status_fields)
```

sourcemash/api/feeds.py
```diff
@@ -24,25 +24,28 @@ def output(self, key, feed):
...
+
+feed_fields_detailed = {
     'description': fields.String,
     'image_url': fields.String,
-    'last_updated': fields.DateTime,
+    'last_updated': fields.DateTime
+}
+feed_fields_detailed.update(feed_fields)
+
+feed_fields_status = {
```
alexgerstein: rename to feed_status_fields?

sourcemash/api/items.py
alexgerstein: Sounds good.

So when a un-logged-in user clicks the saved button, we'll prompt them to sign up, and same for voting?

sourcemash/api/items.py
alexgerstein: Text is full text, right? Summary is what's shown on the card?

tests/test_api/test_items.py
```diff
@@ -17,7 +17,6 @@ def test_get_item_present(self, test_client, item):
 
         data = json.loads(r.data)
         assert data['item']['title'] == item.title
-        assert data['item']['vote'] == 0
```
alexgerstein: Why was this removed?

sourcemash/api/items.py
swglad: Exactly. Just raised this in #165.

sourcemash/api/items.py
swglad: Correct. And we're not ever passing the text to the frontend (nor should we be).

tests/test_api/test_items.py
swglad: Added back. I removed it when I was initially playing with splitting up item_fields before I realized that checking current_user.is_authenticated was the efficiency solution.

sourcemash/api/feeds.py
swglad: Done.

alexgerstein:
I was playing around with the timing on this change and it looks like, even with only our 4 feeds, the typical loading time on the browse page dropped from ~108ms to ~33ms. Nice fix!

PR #168, opened 2015-03-29
Trending backend

### LINKED ISSUE #141: "Trending" items - backend
Trending = absolute value of total votes cast, regardless of direction (e.g. number of upvote clicks + number of downvote clicks; not the same as VoteSum)

swglad:
Closes #141.


migrations/versions/23e65ab635fa_.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '23e65ab635fa'
+down_revision = 'f748d763fc8'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('item', sa.Column('trending', sa.Integer(), nullable=True))
```
alexgerstein: Do we need a whole column for this? Can't we count total votes from our UserItem table pretty quickly?

sourcemash/items/models.py
```diff
@@ -13,6 +13,7 @@ class Item(db.Model):
     category_1 = db.Column(db.String(100))
     category_2 = db.Column(db.String(100))
     voteSum = db.Column(db.Integer, default=0)
+    trending = db.Column(db.Integer, default=0) # total number of votes cast
```
alexgerstein: Do we have an API call to pull out the top X trending items? Like /api/items/trending?

PR #170, opened 2015-03-30
Items trending 2

swglad:


swglad:
Weird frontend issue - See #178 for a better version of thisf same PR

PR #178, opened 2015-04-02
trending items

### LINKED ISSUE #141: "Trending" items - backend

swglad:
Backend only. Closes #141.


swglad:
Travis isn't travis'ing - just a heads up

alexgerstein:
The travis issue should be resolved. Needed to grant permission to the third-party apps we had on the old repo.

sourcemash/api/items.py
```diff
@@ -3,6 +3,9 @@
 from flask import abort
 from flask.ext.restful import Resource, reqparse, inputs, fields, marshal
 from flask.ext.security import current_user, login_required
+from operator import itemgetter
```
alexgerstein: Don't need.

sourcemash/api/items.py
```diff
@@ -77,7 +80,8 @@ def put(self, id):
             user_item = UserItem.query.filter_by(user=current_user, item=item).one()
         except:
             user_item = UserItem(user=current_user, item=item, feed_id=item.feed_id,
-                                 category_1=item.category_1, category_2=item.category_2)
+                                 category_1=item.category_1, category_2=item.category_2,
+                                 last_modified=item.last_updated)
```
alexgerstein: Why is this not set to be datetime.utcnow()?

sourcemash/api/items.py
```diff
@@ -98,6 +103,7 @@ def put(self, id):
         # Toggle saved-for-later status (aka bookmarked)
         if args.saved != None:
             user_item.saved = args.saved
+            user_item.last_modified = datetime.utcnow()
```
alexgerstein: We can probably just call this once since something needs to have changed if the API call was even made. Then we won't need it in the args.vote check too.

sourcemash/api/items.py
```diff
@@ -109,6 +115,26 @@ def get(self):
         user_items = UserItem.query.filter_by(user=current_user, saved=True).all()
         return {'items': [marshal(user_item.item, item_fields) for user_item in user_items]}
 
+class TrendingItemListAPI(Resource):
+
+    @login_required
+    def get(self):
+        trending_user_items = UserItem.query.with_entities(UserItem.item_id, func.count())     \
+                                .filter(UserItem.vote != 0)     \
```
alexgerstein: Would .filter(UserItem.vote) work? Since 0 is None and None is None?

sourcemash/api/items.py
```diff
@@ -109,6 +115,26 @@ def get(self):
...
+                                .filter(UserItem.last_modified > (datetime.utcnow() - timedelta(days=10))) \
```
alexgerstein: Move out the magic number 10 to the top of the file.

sourcemash/api/items.py
```diff
@@ -109,6 +115,26 @@ def get(self):
...
+                                .group_by(UserItem.item_id)        \
+                                .order_by(desc(func.count(UserItem.item_id)))      \
+                                .all()
```
alexgerstein: Is all needed when we have a group_by or order_by?

sourcemash/api/items.py
```diff
@@ -109,6 +115,26 @@ def get(self):
...
+
+        trending_user_items = trending_user_items[:10]
+
+        trending_items = []
+        for item_id, count in trending_user_items:
+             trending_items.append(Item.query.get(item_id))
+
+        return {'items': [marshal(item, item_fields) for item in trending_items]}
```
alexgerstein: Rewrite as:
```python
{'items': [marshal(Item.query.get(id), item_fields) for id, count in trending_items]}
```

sourcemash/api/items.py
```diff
@@ -109,6 +115,26 @@ def get(self):
...
+    @login_required
+    def get(self):
+        trending_user_items = UserItem.query.with_entities(UserItem.item_id, func.count())     \
+                                .filter(UserItem.vote != 0)     \
+                                .filter(UserItem.last_modified > (datetime.utcnow() - timedelta(days=10))) \
+                                .group_by(UserItem.item_id)        \
+                                .order_by(desc(func.count(UserItem.item_id)))      \
+                                .all()
+
+        trending_user_items = trending_user_items[:10]
```
alexgerstein: Rename to trending_items, and you can remove the for loop below.

sourcemash/api/items.py
```diff
@@ -109,6 +115,26 @@ def get(self):
         user_items = UserItem.query.filter_by(user=current_user, saved=True).all()
         return {'items': [marshal(user_item.item, item_fields) for user_item in user_items]}
 
+class TrendingItemListAPI(Resource):
+
+    @login_required
+    def get(self):
+        trending_user_items = UserItem.query.with_entities(UserItem.item_id, func.count())     \
```
alexgerstein: Rename to trending_items [see below].

tests/test_api/test_items.py
```diff
@@ -200,6 +200,29 @@ def test_get_items(self, test_client, user_item_saved):
...
+        self.login(test_client, user_item_upvote.user.email, user_item_upvote.user.password)
+        r = test_client.get('/api/items/trending')
+
+        check_valid_header_type(r.headers)
+        assert r.status_code == 200
+
+        data = json.loads(r.data)
+        assert len(data['items']) == 2
+
+    def test_get_items_missing(self, test_client, user):
```
alexgerstein: Does this test add anything the 'present' test doesn't cover? Do we still have 100% coverage?

sourcemash/api/items.py
swglad: Good catch.

sourcemash/api/items.py
swglad: But do we want it on an unread change too? It would make sense for the answer to that question to be yes.

tests/test_api/test_items.py
swglad: It's not a coverage decider, but the original version was failing at zero items so it's in there for posterity.

tests/test_api/test_items.py
```diff
@@ -200,6 +200,29 @@ def test_get_items(self, test_client, user_item_saved):
...
+        self.login(test_client, user_item_upvote.user.email, user_item_upvote.user.password)
+        r = test_client.get('/api/items/trending')
+
+        check_valid_header_type(r.headers)
+        assert r.status_code == 200
+
+        data = json.loads(r.data)
+        assert len(data['items']) == 1
+
+    def test_get_items_missing(self, test_client, user):
```
alexgerstein: Do you know why the original version was failing on zero items?

sourcemash/api/items.py
```diff
@@ -75,9 +80,12 @@ def put(self, id):
         # Check vote column of user_items table
         try:
             user_item = UserItem.query.filter_by(user=current_user, item=item).one()
+            user_item.last_modified = datetime.utcnow()
+            db.session.commit()
```
alexgerstein: You should be able to do just one commit at the end of the API call, rather than after each change.

tests/test_api/test_items.py
swglad: I was using unzipping to unpack (item, count) tuples. We don't do that anymore. 

Feel free to delete it here and merge if you're ready. Or leave it, either way lol

sourcemash/api/items.py
```diff
@@ -3,11 +3,16 @@
...
 from flask.ext.security import current_user, login_required
+from datetime import datetime, timedelta
+from sqlalchemy import func, desc
 
 from feeds import feed_fields
 from sourcemash.models import Item, UserItem
 from sourcemash.forms import VoteForm
 
+TRENDING_ITEMS_COUNT = 10
+TRENDING_ITEMS_TIMEDELTA = 14
```
alexgerstein: Maybe add ```# Days```?

sourcemash/api/items.py
```diff
@@ -109,6 +117,22 @@ def get(self):
...
+class TrendingItemListAPI(Resource):
+
+    @login_required
+    def get(self):
+        trending_items = UserItem.query.with_entities(UserItem.item_id, func.count())     \
+                                .filter(UserItem.vote)     \
+                                .filter(UserItem.last_modified > (datetime.utcnow() - timedelta(days=TRENDING_ITEMS_TIMEDELTA))) \
+                                .group_by(UserItem.item_id)        \
+                                .order_by(desc(func.count(UserItem.item_id)))      \
+                                .all()
```
alexgerstein: What about the .all? Do we need it?

sourcemash/api/items.py
```diff
@@ -3,11 +3,16 @@
...
 from flask.ext.restful import Resource, reqparse, inputs, fields, marshal
 from flask.ext.security import current_user, login_required
+from datetime import datetime, timedelta
+from sqlalchemy import func, desc
 
 from feeds import feed_fields
 from sourcemash.models import Item, UserItem
 from sourcemash.forms import VoteForm
 
+TRENDING_ITEMS_COUNT = 10
```
alexgerstein: Rename to MAX_TRENDING_ITEMS.

sourcemash/api/items.py
swglad: That didn't work a while back (many many PRs ago), but I don't remember why. I can do that though

sourcemash/api/items.py
swglad: Ah missed that one. It's only the categories.py query as well, but I'll check and remove all of them if it's not necessary.

tests/test_api/test_items.py
alexgerstein: You can delete it when you make the other changes. I don't think we should keep a test for posterity if it's not relevant any more.

sourcemash/api/items.py
```diff
@@ -77,7 +82,8 @@ def put(self, id):
             user_item = UserItem.query.filter_by(user=current_user, item=item).one()
         except:
             user_item = UserItem(user=current_user, item=item, feed_id=item.feed_id,
-                                 category_1=item.category_1, category_2=item.category_2)
+                                 category_1=item.category_1, category_2=item.category_2,
+                                 last_modified=datetime.utcnow())
```
alexgerstein: Remove.

alexgerstein:
Looks great! Just the one small change.

sourcemash/api/items.py
```diff
@@ -109,6 +114,21 @@ def get(self):
         user_items = UserItem.query.filter_by(user=current_user, saved=True).all()
         return {'items': [marshal(user_item.item, item_fields) for user_item in user_items]}
 
+class TrendingItemListAPI(Resource):
+
+    @login_required
+    def get(self):
+        trending_items = UserItem.query.with_entities(UserItem.item_id, func.count())     \
+                                .filter(UserItem.vote)     \
```
alexgerstein: Did you test that this works for both vote == Null and vote == -1/1?

alexgerstein: Nevermind, I see how you did it below. Looks good!

PR #179, opened 2015-04-02
modal on saved click

### LINKED ISSUE #153: Modal subscribe on "Saved" click
Because not just an upvote indicates they may want to subscribe.

swglad:
Closes #153.


sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -69,15 +69,18 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
-            if (this.model.changedAttributes()) {
-            mixpanel.track("Saved", { "Item Title": this.model.get('title'),
-                                        "Feed Title": this.model.feed.get('title') });
-            }
-
         } else {
             this.model.save({saved: true},
                 {success: this.savedToast});
+
+            if (this.model.changedAttributes()) {
```
alexgerstein: Do we need the changedAttributes check anymore?

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -69,15 +69,18 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
+                mixpanel.track("Saved", { "Item Title": this.model.get('title'),
+                                        "Feed Title": this.model.feed.get('title') })
+
+                if (!this.model.feed.get('subscribed')) {
+                    this.showSubscribeModal();
```
alexgerstein: Let's add an option to showSubscribeModal (i.e. {'source': 'saved'}) that adjusts the copy of the subscribe-modal to say either saved or upvoted.

sourcemash/frontend/templates/saved.tpl
```diff
@@ -11,10 +11,10 @@
 <div id="subscribe-modal" class="modal">
   <div class="modal-content">
     <h4>Interested?</h4>
-    <p>You upvoted "<span id="unsubscribed-item-title"></span>" even though you're not subscribed to <i><span id="unsubscribed-feed-title"></span></i>. Would you like to add this feed to your subscriptions?<p>
+    <p>You saved or upvoted "<span id="unsubscribed-item-title"></span>" even though you're not subscribed to <i><span id="unsubscribed-feed-title"></span></i>. Would you like to add this feed to your subscriptions?<p>
```
alexgerstein: Do we need one of these on the saved tpl? Would't they already have clicked save on the items they see here?

sourcemash/frontend/static/js/views/item_card_view.js
swglad: It's on all the other checks (for upvote, etc.). Should it be removed from there too? Could that end up causing other issues?

alexgerstein: Oh yeah. You're right. A user at the moment can click upvote a bunch of times and retrigger the signal. Maybe we should fix the Backbone event to only register the click if it's not active? Then we would be able to remove this.

sourcemash/frontend/static/js/views/item_card_view.js
swglad: Need some help with this

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -31,19 +29,18 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
+        mixpanel.track("Downvoted", { "Item Title": this.model.get('title'),
+                                    "Feed Title": this.model.feed.get('title') })
 	},
 
     voted: function() {
         toast("Vote recorded!", 3000);
     },
 
-    showSubscribeModal: function(e) {
+    showSubscribeModal: function(options, e) {
```
alexgerstein: Remove e.

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -31,19 +29,18 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
         $('#subscribe-modal #unsubscribed-item-title').html(this.model.get('title'));
         $('#subscribe-modal #unsubscribed-feed-title').html(this.model.feed.get('title'));
+        $('#subscribe-modal #unsubscribed-source-title').html(options['source']);
```
alexgerstein: Let's factor subscribed out into a subview.

sourcemash/frontend/templates/category.tpl
```diff
@@ -9,10 +9,10 @@
 <div id="subscribe-modal" class="modal">
   <div class="modal-content">
     <h4>Interested?</h4>
-    <p>You upvoted "<span id="unsubscribed-item-title"></span>" even though you're not subscribed to <i><span id="unsubscribed-feed-title"></span></i>. Would you like to add this feed to your subscriptions?<p>
+    <p>You <span id="unsubscribed-source-title"></span> "<span id="unsubscribed-item-title"></span>" even though you're not subscribed to <i><span id="unsubscribed-feed-title"></span></i>. Would you like to add this feed to your subscriptions?<p>
```
alexgerstein: When this becomes its own template, we can just do ```{{source}}``` here.

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -31,19 +29,18 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
 
     voted: function() {
         toast("Vote recorded!", 3000);
     },
 
-    showSubscribeModal: function(e) {
+    showSubscribeModal: function(options) {
         $('#subscribe-modal #unsubscribed-item-title').html(this.model.get('title'));
         $('#subscribe-modal #unsubscribed-feed-title').html(this.model.feed.get('title'));
+        $('#subscribe-modal #unsubscribed-source-title').html(options['source']);
```
alexgerstein: Sorry to do this to you, but:

would it make more sense to leave event and use a tag in the html to store the source? Then we could access the source like this: http://stackoverflow.com/questions/7823556/passing-parameters-into-the-backbone-events-object-of-a-backbone-view

alexgerstein: Actually scratch that.

sourcemash/frontend/templates/subscribe-modal.tpl
```diff
@@ -0,0 +1,8 @@
+<div class="modal-content">
+  <h4>Interested?</h4>
+  <p>You <span id="unsubscribed-source-title"></span> "<span id="unsubscribed-item-title"></span>" even though you're not subscribed to <i><span id="unsubscribed-feed-title"></span></i>. Would you like to add this feed to your subscriptions?<p>
```
alexgerstein: Do we have quotation issues here? Why is the span for unsubscribed-item-title in quotes?

PR #187, opened 2015-04-06
Mark all as unread.

### LINKED ISSUE #148: Mark all as read
Give user a button on the feed and item page that they can click and mark all of the items on the page as read.

alexgerstein:
Closes #148.


sourcemash/frontend/templates/category.tpl
```diff
@@ -1,4 +1,10 @@
-<h2><%= model.get('category') %></h2>
+<div class="row valign-wrapper">
+    <h2 class="col s10 valign"><%= model.get('category') %></h2>
+
+    <div class="col s2">
+      <button type="submit" class="btn waves-effect waves-light mark-all-read"><i class="mdi-action-done-all"></i></button>
```
swglad: Instead of copying this code twice in category.tpl and feed.tpl, do you want to add it to subscribe-switch-view and make it part of that template (and rename it appropriately, of course!)?

PR #191, opened 2015-04-08
Login errors

### LINKED ISSUE #39: Username/Password errors indicated by where the errors appear.
If you type in an incorrect password, the error "Invalid username/password combination" will show up below the password field.

However, if you type an incorrect username, it will show up below the email field.

So a hacker could figure out valid email addresses in our DB until we get the error messages to always appear in one place.

swglad:
Resolves #39.


swglad:
Play with it first, but I think it's looking good.

sourcemash/frontend/frontend.py
```diff
@@ -1,9 +1,13 @@
...
-from flask import render_template
+from flask import render_template, url_for
 
 
 @bp.route('/')
 @bp.route('/index')
 def index():
     return render_template('index.html')
+
+@bp.route('/login')
```
alexgerstein: Do we still need this?

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -9,12 +9,72 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
     var content = this.template({active: activeTab, current_user: this.user, feeds: this.feeds, categories: this.categories})
-    $('#nav-mobile').html(content);
+    this.$el.html(content);
 
     $('ul.tabs').tabs();
+    return this;
   },
 });
+
+// Source: http://stackoverflow.com/questions/1184624/convert-form-data-to-js-object-with-jquery
```
alexgerstein: Can we move this to sourcemash, or some other generalizable file that's added everywhere?

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -9,12 +9,72 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
     this.listenTo(this.categories, 'sync change:unread_count', this.render);
   },
 
+  events: {
+    'click #login-submit': 'loginSubmit',
```
alexgerstein: Can we make this 'submit #login'?

sourcemash/frontend/templates/sidenav.tpl
```diff
@@ -56,26 +56,24 @@
...
         <input class="validate" id="password" name="password" type="password">
         <label for="password">Password</label>
-        <div class="errors"></div>
+        <div class="errors" id="login-errors"></div>
       </div>
 
       <div class="row">
         <div class="input-field col s5">
-          <button type="submit" class="btn waves-effect waves-light">Login</button>
+          <button type="submit" id="login-submit" class="btn waves-effect waves-light">Login</button>
```
alexgerstein: Should be able to remove login-submit due to above comment.

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -9,12 +9,72 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
+  loginSubmit: function(e){
+    e.preventDefault()
+
+    var formData = JSON.stringify($("#login").serializeObject());
+
+    $.ajax({
+      type: "POST",
+      url: "/login",
+      data: formData,
+      success: function(response){
```
alexgerstein: Rename response to data so we don't have response.response

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -9,12 +9,72 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
+        var user = response.response.user;
+        if (user) {
+          location.reload();
+        }
+        var errors = response.response.errors;
+        if (errors) {
+          errorMsg = errors.email || errors.password || errors.rememeber || {};
```
alexgerstein: var

alexgerstein: I don't think we'l ever have errors on remember.

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -9,12 +9,72 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
+          $("#login-errors").html(errorMsg);
```
alexgerstein: Do we want this to be ```.text(errorMsg)```?

sourcemash/frontend/frontend.py
swglad: Yes, otherwise you ```/login?email=&password=``` with the variables in the url line.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: No. Submit does a post and auto-refresh of the page so we'll never actually render the errors we catch in the JSON. We're overwriting the ability to do a submit by making it a click.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: I agree that we won't, but Flask-Security can return errors on remember, so there's no reason not to catch them here since it costs us nothing.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: It renders identically and we use ```.html()``` everywhere else on the frontend, so I think we should keep it as is.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: I've renamed it ```_serializeObject``` but don't think we should move it to a generalizable file until we use it elsewhere. This is the only place it's needed for now so I'd avoid making it global until there's another application.

sourcemash/frontend/templates/sidenav.tpl
swglad: See above response.

sourcemash/frontend/static/js/views/sidenav_view.js
alexgerstein: The submit happens regardless; what prevents it is the e.preventDefault(). If you make this ```submit #login```, it should still work and you won't need the other IDs in the tpl file.

sourcemash/frontend/frontend.py
alexgerstein: I thought this wasn't overriding the flask-security route though?

swglad: It's not. It makes a post to /login which is Flask-Security, but the way it processes the HTML submit is that is adds the variables to the URL which we don't want since we're not allowing the page to refresh until the login credentials are correct.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: It puts variables in the URL if I make it submit and does not if I make it click.

sourcemash/frontend/static/js/views/sidenav_view.js
alexgerstein: We're overriding jQuery, so if a page includes "sidenav_view.js," jQuery is changed for all other js files. Since we're overriding the global jQuery object, it's weird to only keep it in this one file. That's why I think we should either move the function or remove jQuery.fn from the name.

sourcemash/frontend/frontend.py
alexgerstein: I thought it catches the click?

sourcemash/frontend/static/js/views/sidenav_view.js
alexgerstein: oh. You've tried since we fixed the render el issue?

sourcemash/frontend/frontend.py
swglad: We do catch the click.

However, submit on this login form builds the URL ```.../login``` with variables, which we want to redirect back to index. We want this: (1) so it doesn't show the variables, and (2) because we don't want the login_user.html default template.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: Yes, tested this just now when you asked before responding. I tested everything you asked before commenting.

sourcemash/frontend/frontend.py
alexgerstein: I'm playing around with the branch locally and I don't think it's these lines that are causing either 1 or 2. The template is set in the default config file and the url it's hitting is in flask-security. Still can't figure out why the submit isn't working though.

alexgerstein:
Mixpanel's track_forms was causing another submit. I resolved the issue by moving it out to the function in our backbone.

sourcemash/frontend/frontend.py
```diff
@@ -1,6 +1,6 @@
 from . import bp
 
-from flask import render_template
+from flask import render_template, url_for
```
alexgerstein: Don't need url_for anymore.

sourcemash/frontend/static/js/sourcemash.js
```diff
@@ -26,3 +29,35 @@ Backbone.sync = function(method, model, options) {
 $(document).ready(function() {
     Sourcemash.initialize();
 });
+
+//
+// jQuery
+//
+// Source: http://stackoverflow.com/questions/1184624/convert-form-data-to-js-object-with-jquery
+jQuery.fn._serializeObject = function() {
```
alexgerstein: It's not "private," so i don't think we need the underscore.

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -9,12 +9,43 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
+                      success: this.showErrors,
+                      contentType: "application/json"
+                  });
+  },
+
+  showErrors: function(data) {
+    var user = data.response.user;
+    if (user) {
+      location.reload();
+      mixpanel.track("Logged In");
```
alexgerstein: This may need to come before the location.reload? Not positive.

PR #193, opened 2015-04-11
Better style mark-all-as-read.

### LINKED ISSUE #190: Change mark all as read logo
To mdi-content-drafts. Current logo is very unclear.

alexgerstein:
Closes #190.


swglad:
Looks great. Two quick asks:
1. Can you add some padding around the Mark-Read button and Subscribe switch on the Feed page? They look a bit smashed together.
2. Is the Mark-Read button aligned corrected on the Category page? It's not aligned with the right of the right-most item.

alexgerstein:
I fixed the spacing on mobile. As for the alignment on category pages, I think it should work if you do a hard reset in your browser.

sourcemash/frontend/templates/feed.tpl
```diff
@@ -1,10 +1,8 @@
 <div class="row valign-wrapper">
-    <h2 class="col s8 valign"><%= model.get('title') %></h2>
+    <h2 class="col s8 m9"><%= model.get('title') %></h2>
 
-    <div class="col s2">
-      <button type="submit" class="btn waves-effect waves-light mark-all-read"><i class="mdi-action-done-all"></i></button>
-    </div>
-    <div class="switch subscribe-switch col s2 valign-wrapper"></div>
+    <span title="Mark as Read" class="mark-all-read col s2 m1"><i class="medium mdi-action-done-all"></i></span>
+    <span class="switch subscribe-switch col s2 valign"></span>
```
swglad: Should this also have ```m2``` so that it adds up to 12? I imagine that ```s2``` does that by default since you don't have it, but seems weird not to have it when both ```col``` classes above this line have it.

alexgerstein: s applies to s and everything bigger, so we don't need it.

PR #195, opened 2015-04-14
Scrape one feed: redis queue

### LINKED ISSUE #112: Scrape feed as soon as someone adds it.
Don't need categories right away, but it would be good for users if their feeds get populated quickly.

swglad:
Resolves #112.


swglad:
1. POST feed request 
2. Scrape every 30 mins

swglad:
I removed the function categorize_articles_only() since I figured it's wasteful to queue it and instead we should just wait for the next scrape_and_categorize_articles() that occurs within 30 mins.

heroku.sh
```diff
@@ -1,2 +1,3 @@
 gunicorn sourcemash:create_app\(\) --daemon
-python manage.py scrape
\ No newline at end of file
+python manage.py worker
```
alexgerstein: Isn't this going to block scrape from ever happening?

swglad: Does it need a ```&```?

manage.py
```diff
@@ -46,10 +53,24 @@ def test(all=False):
 def scrape():
     """Start an infinte loop to scrape & categorize articles."""
 
+    q = Queue('default', connection=conn)
     while True:
         logging.info("Starting scrape...")
-        scrape_and_categorize_articles()
-        logging.info("Finished scrape. Let's run it back...")
+        job = q.enqueue(scrape_and_categorize_articles)
```
alexgerstein: Instead of making the full scrape one task, it might help to break it down into feed scrapes and item categorizations. This way, when we add an urgent scrape to the front of the queue, it can be completed sooner.

requirements.txt
```diff
@@ -24,6 +6,7 @@ beautifulsoup4==4.3.2
 billiard==3.3.0.19
 blinker==1.3
 chardet==2.3.0
+click==4.0
```
alexgerstein: What's this?

swglad: Installed with rq

worker/__init__.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+
+def create_worker(env=None):
+
+  import redis
+
+  redis_url = os.getenv('REDISTOGO_URL', 'redis://localhost:6379')
+
```
alexgerstein: Get rid of double-spacing.

sourcemash/api/feeds.py
```diff
@@ -10,6 +10,13 @@
 
 from sourcemash.models import Feed, UserItem, Item
 
+import os
+from rq import Queue
+from worker import create_worker
+from worker.scraper import scrape_articles_only
+
+conn = create_worker(os.environ.get("APP_CONFIG_FILE") or "development")
```
alexgerstein: Is there a way to move conn to the app's init or config?

sourcemash/api/feeds.py
```diff
@@ -81,6 +88,13 @@ def post(self):
             db.session.add(feed)
             db.session.commit()
 
+            # Scrape feed (but don't fail if redis-server is down)
+            try:
+                q = Queue('default', connection=conn)
+                job = q.enqueue_call(func=scrape_articles_only, args=(feed.id,), at_front=True)
```
alexgerstein: Will this work as ```q.enqueue(scrape_articles_only, feed.id, at_front=True)```?

manage.py
swglad: I don’t think it matters much. The limiting factor is really categorization and once that starts, it more than trumps the time to scrape anyway. Scraping 10 new feeds from scratch only took 3.5mins, but categorization (which I stopped after 4 mins) would take much longer that you’d have to wait for it anyway.

sourcemash/api/feeds.py
```diff
@@ -81,6 +88,13 @@ def post(self):
...
+            except:
+                pass
```
alexgerstein: So what happens if the redis server is down? We just wait until the major scrape?

worker/scraper.py
```diff
@@ -47,6 +47,22 @@ def scrape_and_categorize_articles():
         logger.info("CATEGORIZED [%s]: (%s, %s)" % (item.title, item.category_1, item.category_2))
 
 
+def scrape_articles_only(feed_id):
```
alexgerstein: Rename to scrape_feed_articles.

sourcemash/api/feeds.py
swglad: All the examples online with options use enqueue_call() instead of enqueue(), but I can give it a shot if you’d like.

sourcemash/api/feeds.py
swglad: Major scrape wouldn’t even happen I guess. Everything would fail since everything uses the queue… Probably should put a try block around the manage.py scrape too. 

Effectively, if redis is down, so are we.

heroku.sh
alexgerstein: I think so.

manage.py
alexgerstein: So then we should at least separate the categorization tasks, no?

requirements.txt
alexgerstein: ok

sourcemash/api/feeds.py
alexgerstein: no need. I was just curious now that we have it working.

sourcemash/api/feeds.py
alexgerstein: What would happen without the except block?

manage.py
swglad: Doesn’t seem necessary. It requires the option ```depends_on```=scrape_job_id, and the job won’t become enqueued until scrape finishes anyway (how rq works), so effectively it’s just the big function anyway. Might as well wait for the big scrape & categorization then.

sourcemash/api/feeds.py
swglad: I tested it. If the server is down, it crashes and burns with errors to the console. “Connection error 61, etc etc"

sourcemash/api/feeds.py
swglad: It's in the worker's config and I'm not sure if we want to move it into sourcemash's config too. Thoughts?

manage.py
alexgerstein: I'm not sure I'm following. Can you please elaborate?

sourcemash/api/feeds.py
alexgerstein: The app crashes, or just worker?

sourcemash/api/feeds.py
alexgerstein: I think it would make sense to put in sourcemash's config if possible. It's odd to need os and "APP_CONFIG_FILE" in an api file.

sourcemash/api/feeds.py
swglad: Server crashed b/c API call couldn’t finish b/c there was no connection. Worker had already crashed, but server attempted to keep going and crashed here.

alexgerstein: Sweet.

manage.py
swglad: You can’t categorize articles that aren’t there, and rq *does* attempt to fork tasks and complete them simultaneously (so it’s not strictly a queue, unless you add the option ```depends_on``` to the enqueue() call).

sourcemash/api/feeds.py
```diff
@@ -10,6 +10,13 @@
 
 from sourcemash.models import Feed, UserItem, Item
 
+import os
+from rq import Queue
+from worker import create_worker
+from worker.scraper import scrape_feed_articles
+
+conn = create_worker(os.environ.get("APP_CONFIG_FILE") or "development")
```
swglad: What would the process of moving this to a config file look like? Like what lines would go where? I'll do it but I'm pulling a bit of a blank on where to put things now

manage.py
alexgerstein: I was thinking the one task we add ourselves to the queue is scrape articles, then as each article is pulled from the feed, we enqueue a categorization task for the one item.

Our old categorization relied on scraping and categorization happening in that order. If we enqueue after a scrape, then we don't need to worry about ```depends_on```.

sourcemash/api/feeds.py
swglad: Nvm think i got this.

manage.py
swglad: I think we should raise that as a second issue as a more efficient way to do this. Don’t think it strictly needs to be part of this PR (given that there will almost certainly be rq bugs related to that). Can you raise that?

sourcemash/api/feeds.py
alexgerstein: This might simplify a lot of things: https://github.com/mattupstate/flask-rq

swglad: I put this line in sourcemash __init__. Is that right or wrong? 

It doesn't seem that from sourcemash import conn works.

swglad: Okay, I'll look at that repo tomorrow then. Can't do it tonight. Fever coming back and dying here.

manage.py
alexgerstein: Sure. We'll also want to enqueue emails. That's why I think this is a good thing to do.

swglad: Cool

sourcemash/api/feeds.py
swglad: I feel like a moron but I don't know how to do this. Help?

alexgerstein: Let's just not pass a config!

manage.py
```diff
@@ -7,24 +7,31 @@
 from flask.ext.script import Manager, Shell, Server
 from flask.ext.migrate import MigrateCommand
 from flask.ext.assets import ManageAssets
+from rq import Worker, Queue, Connection
+
```
alexgerstein: Remove extra space.

manage.py
```diff
@@ -7,24 +7,31 @@
...
 app = create_app(os.environ.get("APP_CONFIG_FILE") or "development")
+conn = create_worker(os.environ.get("APP_CONFIG_FILE") or "development")
+
 manager = Manager(app)
 
 TEST_CMD = "py.test --cov-report term-missing --cov-config .coveragerc --cov . \
                     --boxed -n14 -k 'not functional' tests/"
 FUNCTIONAL_TEST_CMD = "./functional_test.sh"
 
+THIRTY_MINUTES = 0.5 * 60 * 60
```
alexgerstein: Can we make this ```60 * 60 / 2```?

manage.py
```diff
@@ -46,10 +53,24 @@ def test(all=False):
 def scrape():
     """Start an infinte loop to scrape & categorize articles."""
 
+    q = Queue('default', connection=conn)
     while True:
         logging.info("Starting scrape...")
```
alexgerstein: This should be logger.info("Starting scrape..."). See how it's done in scraper.

manage.py
```diff
@@ -46,10 +53,24 @@ def test(all=False):
...
-        scrape_and_categorize_articles()
-        logging.info("Finished scrape. Let's run it back...")
+        job = q.enqueue_call(func=scrape_and_categorize_articles, timeout=1800)
+        logging.info("Enqueued scrape. Let's run it back in 30 mins...")
```
alexgerstein: Same here.

worker/__init__.py
```diff
@@ -0,0 +1,10 @@
+import os
```
alexgerstein: Move inside create_worker.

worker/scraper.py
```diff
@@ -47,6 +47,22 @@ def scrape_and_categorize_articles():
         logger.info("CATEGORIZED [%s]: (%s, %s)" % (item.title, item.category_1, item.category_2))
 
 
+def scrape_feed_articles(feed_id):
+    feed = Feed.query.get(feed_id)
```
alexgerstein: Now that the queue is working, can we see if there's a way to pass the feed object? It would be nice to not have to do a lookup.

PR #202, opened 2015-04-19
Make categories a table.

### LINKED ISSUE #110: Category table: change api, db, frontend, backend

alexgerstein:
Closes #110.


swglad:
Check that adding feed bug we saw during the meeting today.

migrations/env.py
```diff
@@ -57,7 +57,8 @@ def run_migrations_online():
     connection = engine.connect()
     context.configure(
                 connection=connection,
-                target_metadata=target_metadata
+                target_metadata=target_metadata,
+                render_as_batch=True
```
swglad: Did you change this?

sourcemash/api/categories.py
```diff
@@ -16,59 +17,50 @@ def get(self):
...
-        distinct_category_1 = Item.query.with_entities(Item.category_1, func.count())   \
-                                        .filter(Item.feed_id.in_(user_feed_ids))        \
-                                        .group_by(Item.category_1)                      \
-                                        .all()
-        distinct_category_2 = Item.query.with_entities(Item.category_2, func.count())   \
-                                        .filter(Item.feed_id.in_(user_feed_ids))        \
-                                        .group_by(Item.category_2)                      \
-                                        .all()
+        categories = Category.query \
+                             .with_entities(Category.id,
```
swglad: Do you need ```Category.id```? Doesn't look like you do.

sourcemash/categories/models.py
```diff
@@ -0,0 +1,13 @@
+from sqlalchemy.ext.associationproxy import association_proxy
```
swglad: Necessary?

sourcemash/frontend/templates/item-card.tpl
```diff
@@ -15,8 +15,11 @@
       </div>
 
       <div class="card-action valign-wrapper row">
-        <div class="col s4"><a href="/#categories/<%=item.get('category_1')%>"><%=item.get('category_1')%></a></div>
-        <div class="col s4"><a href="/#categories/<%=item.get('category_2')%>"><%=item.get('category_2')%></a></div>
+        <div class="category-list col s8">
```
swglad: Can you try making this a ```<p class="flow-text">...</p>``` instead? The text is very heavy now that it's a list.

tests/fixtures/app_fixtures.py
```diff
@@ -5,8 +5,7 @@
 from sqlalchemy.orm import Session, scoped_session, sessionmaker
 from sourcemash import create_app
 from sourcemash.database import db as _db
-from tests.factories import (feed_factories, item_factories, role_factories, user_factories,
-                            user_item_factories)
+from tests.factories import (feed_factories, item_factories, role_factories, user_factories, user_item_factories)
```
swglad: Why are the imports in parentheses?

tests/test_api/test_items.py
```diff
@@ -228,15 +228,13 @@ def test_get_items(self, test_client, item):
...
         check_valid_header_type(r.headers)
-        assert r.status_code == 200
-
-        data = json.loads(r.data)
-        assert len(data['items']) == 0
+        assert r.status_code == 404
 
     def test_get_category_items_present(self, test_client, itemsWithCategory):
-        r = test_client.get('/api/categories/' + itemsWithCategory[0].category_1 + '/items/all')
+        category = itemsWithCategory[0].cats.first()
```
swglad: .cats and not .categories?

tests/test_api/test_items.py
```diff
@@ -251,22 +249,9 @@ def test_get_items_present(self, test_client, userWithPopulatedFeed):
         self.login(test_client, userWithPopulatedFeed.email, userWithPopulatedFeed.password)
 
         feed = userWithPopulatedFeed.subscribed.first()
+        category = feed.items[0].cats.first()
```
swglad: .cats again?

tests/test_api/test_items.py
```diff
@@ -228,15 +228,13 @@ def test_get_items(self, test_client, item):
 class TestCategoryItemListAllAPI:
 
     def test_get_category_items_missing_category(self, test_client):
-        r = test_client.get('/api/categories/nonexistent_category/items/all')
+        r = test_client.get('/api/categories/100/items/all')
```
swglad: Does /all here make sense?

tests/test_api/test_items.py
```diff
@@ -294,8 +274,9 @@ def test_get_items_category_with_unsubscribed_item(self, test_client, userWithPo
 
         feed = userWithPopulatedFeed.subscribed.first()
         user_items_length = feed.items.count()
+        category = feed.items[0].cats.first()
```
swglad: .cats

worker/categorize.py
```diff
@@ -387,8 +376,7 @@ def _get_best_keywords(self, communities, original_keywords):
             best_original_keywords = keyword_counts.most_common(2)
             best_keywords.update([keyword[0] for keyword in best_original_keywords])
 
-        return [keyword.title() for keyword in best_keywords] if best_keywords else [""]
-
+        return [keyword.title() for keyword in best_keywords]
```
swglad: How were you able to drop the if?  What if best_keywords is empty?

worker/scraper.py
```diff
@@ -37,14 +38,14 @@ def scrape_and_categorize_articles():
...
+            item.categories.append(category)
 
-        if len(categories) >= 2:
-            item.category_2 = categories[1]
+        item.categorized = True
 
         db.session.commit()
-        logger.info("CATEGORIZED [%s]: (%s, %s)" % (item.title, item.category_1, item.category_2))
+
+        logger.info("CATEGORIZED [%s]" % (item.title))
```
swglad: Can you print the categories too? Perhaps do a ```map```/```.join``` on the list of category strings and then add them to the logger? I think it's useful information to render to the log and is worth the bit of computational effort.

migrations/env.py
alexgerstein: Yeah. It turns out flask-migrate did not "fix itself." I kept running into issues trying to drop columns. I think what you tested for that issue must have been just adding columns, which works fine with sqlite. Render_as_batch writes the migration script as a batch migration. If the migration occurs on a sqlite table, it makes a new table with the modified columns and then copies over the data. If it's on another db, it just processes each step as usual. You can read more about it here: http://alembic.readthedocs.org/en/latest/batch.html

sourcemash/api/categories.py
alexgerstein: I need it because we are passing id in the API call.

sourcemash/categories/models.py
alexgerstein: Nope. I'll remove.

sourcemash/frontend/templates/item-card.tpl
alexgerstein: What will flow-text do?

tests/fixtures/app_fixtures.py
alexgerstein: No reason. They just always have been. Should I take them out?

tests/test_api/test_items.py
alexgerstein: Yeah because that's what we're testing: "TestCategoryItemListAllAPI".

tests/test_api/test_items.py
alexgerstein: .cats is the actual association. .categories is just the association proxy.

Association proxies just work as a convenience for when you're constructing/modifying models. To actually pull out the category models, rather than just their unicode representation (as strings), we need to use cats.

tests/test_api/test_items.py
alexgerstein: See above.

tests/test_api/test_items.py
alexgerstein: See above.

worker/categorize.py
alexgerstein: It just won't go into a for loop when appending categories. Since I check for .categorized now instead, we don't need to create a NULL category anymore.

worker/scraper.py
alexgerstein: Sure. I was assuming the lists would be annoyingly long when I first wrote it.

migrations/env.py
swglad: Cool, thanks for figuring that.

sourcemash/api/categories.py
swglad: Really? Where?

tests/test_api/test_items.py
swglad: My bad, all good.

tests/test_api/test_items.py
swglad: Ahh okay, thanks

sourcemash/api/categories.py
alexgerstein: See line 51 where we pass the data back to the user.

worker/categorize.py
swglad: Sounds good. Can you point me to the .categorized check in the code when you get a chance?

sourcemash/frontend/templates/item-card.tpl
```diff
@@ -15,8 +15,11 @@
       </div>
 
       <div class="card-action valign-wrapper row">
-        <div class="col s4"><a href="/#categories/<%=item.get('category_1')%>"><%=item.get('category_1')%></a></div>
-        <div class="col s4"><a href="/#categories/<%=item.get('category_2')%>"><%=item.get('category_2')%></a></div>
+        <div class="category-list flow-text col s8">
```
swglad: flow-text only works as a class on a <p>. It's what makes the text responsive (see browse feed page)

worker/categorize.py
alexgerstein: Line 44 in worker/scraper.py

sourcemash/api/categories.py
swglad: Cool

worker/categorize.py
swglad: Perfect, thank you

PR #219, opened 2015-04-22
Loading

### LINKED ISSUE #147: Loading Bar
Categories (and now feeds too) take a while to calculate unread counts. Let's add a loading bar/circle so that the user knows the lists are on their way.

alexgerstein:
Closes #147.


sourcemash/frontend/static/js/routers/app_router.js
```diff
@@ -70,10 +71,15 @@ Sourcemash.Routers.AppRouter = Backbone.Router.extend({
         var savedItems = new Sourcemash.Collections.Items([], {saved: true});
         var savedView = new Sourcemash.Views.SavedView({ collection: savedItems });
 
-        savedView.collection.fetch({feeds: this._feeds, categories: this._categories, success: function() {savedView.render()}});
+        savedView.collection.fetch({feeds: this._feeds, categories: this._categories, success: _.bind(function() {this._loaded(savedView)}, this)});
         this._swapView(savedView);
     },
 
+    _loaded: function(view) {
```
swglad: Why do we need to bind this function to all the views (as above)?

sourcemash/api/categories.py
```diff
@@ -7,6 +7,12 @@
 from sqlalchemy import func
 from collections import Counter
 
+class CategoryAPI(Resource):
```
swglad: Why did you make this API call / change from "model" to "title" as part of this PR? Just curious.

PR #221, opened 2015-04-23
Modal sign up

### LINKED ISSUE #165: Modal to "Sign-up" for un-logged in users

### LINKED ISSUE #226: Forgot Password view
With JSON form, like register/login form.

swglad:
Resolves #165. Closes #226.


config/default.py
```diff
@@ -16,6 +16,7 @@
 SECURITY_PASSWORD_HASH = "bcrypt"
 SECURITY_PASSWORD_SALT = "abcde1234509876zyxwvu22"
 SECURITY_LOGIN_USER_TEMPLATE = "index.html"
+SECURITY_REGISTER_USER_TEMPLATE = "index.html"
```
alexgerstein: Does the modal pop up if you click, "I don't have an account." in the sidenav? Can we do that instead of redirecting to a new page?

config/development.py
```diff
@@ -8,6 +8,8 @@
 MAIL_USE_TLS = False
 MAIL_USE_SSL = True
 
+SECURITY_SEND_REGISTER_EMAIL = False
```
alexgerstein: Do we still need this?

sourcemash/frontend/assets.py
```diff
@@ -39,8 +39,10 @@
                 "js/views/feed_view.js",
                 "js/views/saved_view.js",
                 "js/views/browse_view.js",
+                "js/views/register_view.js",
```
alexgerstein: Do we have a login_view?

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -40,10 +47,16 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
     this.loadingView = new Sourcemash.Views.LoadingView({loading: this.loading});
     this.$(".loading").html(this.loadingView.render().el);
 
+    // Render register modal view
+    this.registerModalView = new Sourcemash.Views.RegisterModalView();
+    this.$("#register-modal").html(this.registerModalView.render().el);
+
     // Render item cards
     var feedCards = [];
+    _user = this.user;
```
alexgerstein: If you do a ```_.bind``` on the function below, you don't need to save this.user to _user.

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -44,7 +53,21 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
+    },
+
+    showRegisterModal: function(options) {
+        $('#register-modal').openModal();
+
+        mixpanel.track("Register Modal", { "Item Title": this.model.get('title'),
+                                            "Feed Title": this.model.feed.get('title') });
+    },
+
+    showRegisterModal: function(options) {
```
alexgerstein: Is there a reason we have this twice?

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -70,16 +93,20 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
-                {success: this.savedToast});
-
-            mixpanel.track("Saved", { "Item Title": this.model.get('title'),
-                                    "Feed Title": this.model.feed.get('title') })
-
-            if (!this.model.feed.get('subscribed')) {
-                this.showSubscribeModal({'source':'saved'});
-            }
-        }
+            if (!this.user.get('id')) {
```
alexgerstein: It really sucks having this check everywhere.

Would it be possible to have the modal in the same place in the html for every page (i.e. like the side-nav)? Maybe we could add a global error interceptor (like for errors) that launches the pages modal whenever the status code is a 401 (unauthorized) error.

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,41 @@
...
+    registerFromModal: function(e) {
+        e.preventDefault();
+        var formData = JSON.stringify($("#register").serializeObject());
+        $('#register-password').val('');
+        $('#register-password-confirm').val('');
+        var posting = $.ajax({
+                          type: "POST",
+                          url: "/register",
+                          data: formData,
+                          success: this.showErrors,
```
alexgerstein: On success showErrors? will errors ever appear? Also, shouldn't the errors automatically populate because of the code we wrote in default.js?

config/default.py
swglad: All this does is get rid of "/register" as a page you can go to by making it just be the index page. Same exact thing we did to login.

swglad: No, the modal doesn't pop up when you click "I don't have an account." It can't be done easily, and I think it's actually cleaner to see Login in the sidenav and Register in the view if you click on that.

config/development.py
swglad: Nope, guess not.

sourcemash/frontend/assets.py
swglad: No, the login form is strictly part of the sidenav. Register view is only nice b/c then it can be seen side-by-side with the login form in the sidenav.

sourcemash/frontend/static/js/views/item_card_view.js
swglad: Rebase mess up. Nice catch.

sourcemash/frontend/static/js/views/item_card_view.js
swglad: You can't put the modal in the side nav b/c it won't load in front of the whole page if you do. It will only load in the sidenav, which means it get's cut off when you enter the "body" section of the page.

swglad: Doesn't matter if there's a global interceptor, the modal has to be there, unless you put it in base.html, which I don't think is a great idea.

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: I'll rename the function. It really is registerOrShowErrors.

swglad: And no, the errors need to be extracted from the JSON response and directly rendered to the modal if we want to see them. 

The stuff we added last time allows the JSON message to be constructed before being sent, not parsed when it's received.

sourcemash/frontend/static/js/views/browse_view.js
swglad: Would that require writing it as it's own function and calling it here then? Is that necessary?

alexgerstein: No. Just explaining how bind is useful, and how it's another way to do things.

sourcemash/frontend/static/js/views/item_card_view.js
alexgerstein: I didn't mean putting it in sidenav, but rather putting it in it's own view that gets rendered to the base.html like how we render sidenav.

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: Is this how we do it for the login errors too?

sourcemash/frontend/static/js/views/browse_view.js
swglad: Cool, much appreciated

sourcemash/frontend/static/js/views/item_card_view.js
swglad: Not sure if I love the idea of keeping the view at the "higher" level and having clicks on each smaller view be passed "up" and captured by that global. It could definitely work, but the composition/structure seems non-ideal.

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: Yes. The function is actually copied verbatim (except with different selectors for error rendering in the html). I'll change the signatures on both.

sourcemash/frontend/assets.py
alexgerstein: Didn't we say we wanted login and register to both be on the modal in some way?

sourcemash/frontend/static/js/views/item_card_view.js
alexgerstein: Having this logic in every view that needs an authenticated user still makes me uneasy. I think it's poor decomposition to require that you need to search in 5 different files if we ever want to change how the registration modal is done.

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,41 @@
...
+    registerFromModal: function(e) {
+        e.preventDefault();
+        var formData = JSON.stringify($("#register").serializeObject());
+        $('#register-password').val('');
+        $('#register-password-confirm').val('');
+        var posting = $.ajax({
+                          type: "POST",
+                          url: "/register",
+                          data: formData,
+                          success: this.registerOrShowErrors,
```
alexgerstein: http://api.jquery.com/jquery.ajax/ shows that there is an 'error' option for an ajax call. It would probably make more sense to split the success and error callbacks if possible.

sourcemash/frontend/static/js/views/register_view.js
```diff
@@ -0,0 +1,41 @@
...
+
+    events: {
+        'submit #register': 'register',
+    },
+
+    register: function(e) {
+        e.preventDefault();
+        var formData = JSON.stringify($("#register").serializeObject());
+        $('#register-password').val('');
+        $('#register-password-confirm').val('');
```
alexgerstein: I thought register password confirm doesn't work?

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -21,12 +21,12 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
                       type: "POST",
                       url: "/login",
                       data: formData,
-                      success: this.showErrors,
+                      success: this.loginOrShowErrors,
```
alexgerstein: See above comment about the error option.

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,41 @@
...
+
+    registerOrShowErrors: function(data) {
+        var user = data.response.user;
+        if (user) {
+          mixpanel.track("Register", {"User": user});
+          $("#register-modal").closeModal();
+          toast("Check your email for confirmation!", 2000);
+          setTimeout(function(){window.location.replace("/")}, 2000);
+        };
+        var errors = data.response.errors;
```
alexgerstein: Could we extract this error code to the AJAX catch? We already wrote the logic for rendering errors there, so it can clean up this function a little if we move this there.

sourcemash/frontend/static/js/views/item_card_view.js
swglad: It's not great, but requiring the "pass up" logic seems worse. Is there another option?

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: I'll play with it, but I don't think the request "fails" in that sense. There is no "error" field, but an "errors" entry in the JSON returned. No timeout, abort, or parseError has ever been included in the responses that I've looked at.

sourcemash/frontend/static/js/views/register_view.js
swglad: Good catch, will delete.

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,41 @@
...
+
+    events: {
+        'submit #register': 'registerFromModal',
+    },
+
+    registerFromModal: function(e) {
+        e.preventDefault();
+        var formData = JSON.stringify($("#register").serializeObject());
+        $('#register-password').val('');
+        $('#register-password-confirm').val('');
```
swglad: Delete.

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: There's no 409 status code or anything?

sourcemash/frontend/assets.py
swglad: I thought we ended that conversation with login being in the sidenav and making a register view so they can be side-by-side. And the login is always in the sidenav while you're not logged it, so then you'll see the same form twice on the screen: once in the modal and once in the sidenav. That's weird, no?

alexgerstein: What happens if I'm already a user on the site and accidentally upvote before logging in? The modal pops up to register. Do I need to close the modal first in order to sign in? Maybe we should at least have an "I already have an account link" on the modal which can route to a more explicit login view, like the one you made for register?

sourcemash/frontend/static/js/views/item_card_view.js
alexgerstein: It's not really a "pass up" logic though, because these views that require registration will never know about it. That type of modular decomposition (where one view doesn't know/care about the other views) is ideal, no?

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: Just checked. It gives a generic 400, but that should be good enough to work with.

sourcemash/frontend/assets.py
swglad: But then you'll have a login form next to a login form.....

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: Wait, will that allow us to still render the errors in the correct place on the form?

alexgerstein: Yeah, if you follow the same format as the other ```class="error"``` lines.

sourcemash/frontend/static/js/views/item_card_view.js
swglad: It completely is. Clicks at the item-card, feed-card, feed, and category level would need to "propagate up" and be captured by a higher-order handler.

sourcemash/frontend/assets.py
alexgerstein: ok.

sourcemash/frontend/static/js/views/item_card_view.js
alexgerstein: It's not the click you'd be capturing though, but rather the API call.

sourcemash/frontend/assets.py
swglad: It's not hard to do if you think it's useful, but isn't that design a bit odd? 

Maybe we should add a modal footer that says "Already have an account", and when you click on it, it activates the login form in the sidenav (not sure how that would work with the Backbone views). Would you like to raise an issue for that?

sourcemash/frontend/static/js/views/item_card_view.js
swglad: But there are multiple API calls: PUT on item, POST on feed, PUT on feed (mark-all-as-read)

sourcemash/frontend/assets.py
alexgerstein: I just think i'd be annoyed if the modal pops up and the only action i can do is a "Godammit. Cancel."

sourcemash/frontend/static/js/views/item_card_view.js
alexgerstein: yeah, but you can take the url and send that to mixpanel if you're concerned about tracking the source.

Wouldn't it be nice if whenever we add a new login_required feature that we never need to write code for pulling up the register modal again?

sourcemash/frontend/assets.py
swglad: What do you think of the modal footer idea in the comment above?

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: Yeah that should trigger error.

sourcemash/frontend/assets.py
alexgerstein: it would close the modal and activate the email field?

swglad: Yeah. Can that just be done with selectors?

sourcemash/frontend/static/js/views/item_card_view.js
swglad: So it would require a server response to render the register modal then?

alexgerstein: The modal will always be rendered (like how we render the sidenav). The server response would just trigger a modal open.

sourcemash/frontend/assets.py
alexgerstein: Yeah. I like that idea.

sourcemash/frontend/static/js/views/item_card_view.js
swglad: Not sure how that implementation would work, but we can talk about it for next week then

alexgerstein: In index.html, we have divs for top-nav, side-nav, and main. We'd add another div for register-modal. The registerModal view would be rendered in the initialize function of app_router, just like sidenav is at the moment.

Then, in default.js, we already have a function that is called on ajaxError. Right now it just renders error messages, but we'll add another part that opens the modal if the error code is a 401 Unauthorized Error. If yes, you just call $(".register-modal").openModal() or whatever it is, and send mixpanel an event with the url as the source.

swglad: Kk. I'll do this as my first issue of next week

sourcemash/frontend/static/js/routers/app_router.js
```diff
@@ -37,8 +41,16 @@ Sourcemash.Routers.AppRouter = Backbone.Router.extend({
...
+        if (!this._user.get('id')) {
+            $("#register-modal").openModal();
+        } else {
+            window.location.replace("/");
+        };
+    },
+
     browseFeeds: function() {
-        var browseView = new Sourcemash.Views.BrowseView({collection: this._feeds});
+        var browseView = new Sourcemash.Views.BrowseView({ collection: this._feeds, user: this._user });
```
alexgerstein: I don't think we need to pass user to these views anymore either.

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -42,8 +44,10 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
 
     // Render item cards
     var feedCards = [];
+    _user = this.user;
     this.collection.models.forEach(function(feed) {
-        var feedCardView = new Sourcemash.Views.FeedCardView({el: "#feed-card-" + feed.get('id'), model: feed});
+        var feedCardView = new Sourcemash.Views.FeedCardView({el: "#feed-card-" + feed.get('id'),
+                                                              model: feed, user: _user});
```
alexgerstein: Don't think we need user here either

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -9,8 +9,10 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
     this.typeahead = new ExtendedTypeahead({collection: this.collection, key: 'title'});
     this.listenTo(this.collection, 'sync', this.render);
 
-    this.feedCardViews = [];
     this.loading = true;
+    this.user = options.user;
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/feed_card_view.js
```diff
@@ -2,18 +2,20 @@ Sourcemash.Views.FeedCardView = Backbone.View.extend({
     template: JST['feed-card'],
 
     initialize: function(options) {
+        this.user = options.user;
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/feed_card_view.js
```diff
@@ -2,18 +2,20 @@ Sourcemash.Views.FeedCardView = Backbone.View.extend({
...
         this.listenTo(this.model, 'change', this.render);
         this.render();
     },
 
     render: function() {
         this.$el.html(this.template({ model: this.model }));
-        this.subscribeSwitchView = new Sourcemash.Views.SubscribeSwitchView({ model: this.model });
-        this.$(".subscribe-switch").html(this.subscribeSwitchView.render().el)
+        this.subscribeSwitchView = new Sourcemash.Views.SubscribeSwitchView({ model: this.model,
+                                                                              user: this.user });
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -2,6 +2,7 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
     template: JST['item-card'],
 
     initialize: function(options) {
+        this.user = options.user;
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -4,6 +4,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
             this.listenTo(this.model, 'change:subscribed change:title', this.render);
         };
 
+        this.user = options.user;
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -36,14 +37,17 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
 
         // Render subscribe-toggle switch if feed page
         if (this.model) {
-            this.subscribeSwitchView = new Sourcemash.Views.SubscribeSwitchView({ model: this.model });
+            this.subscribeSwitchView = new Sourcemash.Views.SubscribeSwitchView({ model: this.model,
+                                                                                  user: this.user });
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -36,14 +37,17 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
...
             this.$(".subscribe-switch").html(this.subscribeSwitchView.render().el);
         };
 
         // Render item cards
         var itemCards = [];
+        _user = this.user;
         this.collection.models.forEach(function(item) {
-            var itemCardView = new Sourcemash.Views.ItemCardView({el: "#item-" + item.get('id'), model: item});
+            var itemCardView = new Sourcemash.Views.ItemCardView({el: "#item-" + item.get('id'), model: item,
```
alexgerstein: Or here.

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -36,14 +37,17 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
...
         if (this.model) {
-            this.subscribeSwitchView = new Sourcemash.Views.SubscribeSwitchView({ model: this.model });
+            this.subscribeSwitchView = new Sourcemash.Views.SubscribeSwitchView({ model: this.model,
+                                                                                  user: this.user });
             this.$(".subscribe-switch").html(this.subscribeSwitchView.render().el);
         };
 
         // Render item cards
         var itemCards = [];
+        _user = this.user;
```
alexgerstein: Here.

sourcemash/frontend/static/js/views/register_view.js
```diff
@@ -0,0 +1,41 @@
...
+                      });
+    },
+
+    registerOrShowErrors: function(data) {
+        var user = data.response.user;
+        if (user) {
+          mixpanel.track("Registered");
+          $("#register-modal").closeModal();
+          toast("Check your email for confirmation!", 2000);
+          setTimeout(function(){window.location.replace("/")}, 2000);
```
alexgerstein: Is there a better way for us to do this?

sourcemash/frontend/static/js/views/subscribe_switch_view.js
```diff
@@ -1,7 +1,8 @@
 Sourcemash.Views.SubscribeSwitchView = Backbone.View.extend({
     template: JST['subscribe-switch'],
 
-    initialize: function() {
+    initialize: function(options) {
+        this.user = options.user;
```
alexgerstein: Here.

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,41 @@
...
+                          data: formData,
+                          success: this.registerOrShowErrors,
+                          contentType: "application/json"
+                      });
+    },
+
+    registerOrShowErrors: function(data) {
+        var user = data.response.user;
+        if (user) {
+          mixpanel.track("Registered");
```
alexgerstein: Let's rename to "Created Account"

sourcemash/frontend/static/js/views/register_view.js
```diff
@@ -0,0 +1,41 @@
...
+                          data: formData,
+                          success: this.registerOrShowErrors,
+                          contentType: "application/json"
+                      });
+    },
+
+    registerOrShowErrors: function(data) {
+        var user = data.response.user;
+        if (user) {
+          mixpanel.track("Registered");
```
alexgerstein: Let's rename to "Created Account"

sourcemash/frontend/assets.py
swglad: Done!

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: Doesn't look like this can be done unfortunately. While the packet returned has error codes in the 400's, Flask-Security still sends back a 200 as long as the processing was successful. So we get a 200 whether the packet contains errors or an authenticated user, so we have to leave it as a check-for-both function.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: See above response.

swglad:
BUG: you can't click on "I need an account" more than once. The modal will only pop up the first time. @alexgerstein any chance you can take this one?

swglad:
Nvm got the bug.

sourcemash/api/__init__.py
```diff
@@ -4,6 +4,17 @@
...
 api = Api(bp, prefix="/api")
 
+from functools import wraps
+from flask.ext.security import current_user
+
+def login_required(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        if not current_user.is_authenticated():
+            return {"errors": "Not logged in."}, 401
```
alexgerstein: To keep this consistent with our other errors, can we do ```{"errors": {"user": ["User is not authenticated."]}}, 401```

sourcemash/api/users.py
```diff
@@ -50,12 +49,14 @@ def __init__(self):
     def get(self):
         return { 'user': marshal(current_user, user_fields) }
```
alexgerstein: What does this return if current_user is anon?

swglad: Email: null, id: 0 (both of which don’t exist and pass our HTML/JS checks as expected)

sourcemash/frontend/static/js/routers/app_router.js
```diff
@@ -13,11 +13,17 @@ Sourcemash.Routers.AppRouter = Backbone.Router.extend({
 
         self._user = new Sourcemash.Models.User();
         self._feeds = new Sourcemash.Collections.Feeds([], {allFeeds: true});
-        self._categories = new Sourcemash.Collections.Categories();
+        self._categories = new Sourcemash.Collections.Categories([], {allCategories: true});
```
alexgerstein: This would show all categories in the sidenav, wouldn't it?

swglad: No, we only show categories if the user is logged in. It works exactly the same as feeds did before (and works correctly aka how we thought it was working before this change)

sourcemash/frontend/templates/forgot-modal.tpl
```diff
@@ -0,0 +1,24 @@
...
+          <button type="submit" class="btn waves-effect waves-light">Recover Password</button>
+        </div>
+      </div>
+    </form>
+  </div>
+</div>
+<p class="divider"><p>
+<div class="modal-footer">
+  <a href="#" class="forgot-close waves-effect waves-green btn-flat modal-action modal-close">I remember!</a>
+  <a href="#" class="waves-effect waves-green btn-flat modal-action modal-close">I'll pass...</a>
```
alexgerstein: Do we need this button?

sourcemash/frontend/static/js/views/forgot_modal_view.js
```diff
@@ -0,0 +1,35 @@
...
+                      });
+    },
+
+    recoverPassword: function(data) {
+      $("#forgot-modal").closeModal();
+      toast("Check your email for further instructions.", 3000);
+    },
+
+    activateLoginForm: function() {
+      toast("<-- Login over there!", 3000)
```
alexgerstein: I don't think we need this functionality here because the user decides to click on "forgot password."

sourcemash/frontend/templates/forgot-modal.tpl
swglad: For functionality, no? For consistency and based on the fact that you had kinda mentioned you like the option for a "cancel" button, yes. It does no harm and it's on every other modal, which is the only reason that I included it.

sourcemash/frontend/static/js/views/forgot_modal_view.js
swglad: Up to you. I think the modal looks a little nicer with a footer, and again it doesn't do any harm to leave it there. The reset password form is very small and flat, so these buttons add some form of interaction and interest, even if it's not strictly necessary.

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,46 @@
...
+        };
+        var errors = data.response.errors;
+        if (errors) {
+          errorMsg = errors.email || errors.password || {};
+          $("#register-errors").html(errorMsg);
+        };
+    },
+
+    activateLoginForm: function() {
+      toast("<-- Login over there!", 3000)
```
alexgerstein: Not great UI. Do you think it's enough to just put the login in focus?

sourcemash/frontend/static/scss/sourcemash.scss
```diff
@@ -17,7 +17,7 @@
   display: none !important;
 }
 
-.loading {
+.preloader-wrapper {
```
alexgerstein: Is there a reason we switched this? Just curious.

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: I know, I hate it. But it’s not enough. I barely noticed it working sometimes and I was looking for it…

And you can’t put HTML/icons in toasts, so this was the best I could come up. I don’t like “Log in on the left!” or “Login in the menu bar!"

sourcemash/frontend/templates/register.tpl
```diff
@@ -0,0 +1,21 @@
+<h2>Sign Up</h2>
```
alexgerstein: Do we still use the register page? Or is everything on the modal now?

tests/test_api/test_users.py
```diff
@@ -82,7 +82,3 @@ def test_delete_user(self, test_client, user):
         check_valid_header_type(delete.headers)
         data = json.loads(delete.data)
         assert data['result'] == True
-
-        # Dummy user should no longer be logged in
```
alexgerstein: Shouldn't we check for this?

sourcemash/frontend/static/scss/sourcemash.scss
swglad: Because “hide” was messing with the padding when the user is versus isn’t logged in, so I moved it one div higher.

tests/test_api/test_users.py
swglad: All this checks now is whether login_required is working, which I made as a separate test (above)

sourcemash/api/users.py
alexgerstein: Wouldn't if (user) be true in this case though?

sourcemash/frontend/templates/register.tpl
swglad: Good catch! Will delete file

Scott

sourcemash/frontend/static/js/routers/app_router.js
alexgerstein: How?

sourcemash/frontend/static/js/views/forgot_modal_view.js
alexgerstein: I think we should take it out for this form. We can leave the footer for the submit button though.

sourcemash/api/users.py
swglad: That was always true, which is why we started with the whole .get('email') thing.

sourcemash/frontend/static/js/routers/app_router.js
swglad: See the HTML. It literally works identically to how you wrote feeds in the UI redesign

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: Even a toast doesn't mean that the user will know the field is already highlighted for them though. I don't think it adds much, and it might be confusing to get a toast even though nothing happened on the backend.

sourcemash/frontend/static/scss/sourcemash.scss
alexgerstein: Makes sense.

sourcemash/frontend/templates/forgot-modal.tpl
alexgerstein: Maybe let's just call it "Cancel" or "Not Now/Yet" then.

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: If you play with it, you’ll see it’s confusing where to go. The toast appears because you clicked “I already have an account”, and it’s weird if the user thinks nothing happened except the modal closing when they clicked that

alexgerstein: What if it brings up a login modal?

sourcemash/frontend/static/js/routers/app_router.js
alexgerstein: Feeds can be filtered by subscribed v. unsubscribed. How can we do the same for categories?

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: Then a modal will lead to another modal?

Plus, I like the login form in the sidenav a lot!

sourcemash/api/users.py
alexgerstein: So what happens if your login is incorrect for loginOrShowErrors or registerOrShowErrors?

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: We can keep it there, but if we're concerned it's not explicit enough, maybe we bring up a modal in this instance?

tests/test_api/test_users.py
alexgerstein: Sounds good.

sourcemash/frontend/static/js/routers/app_router.js
swglad: It makes another API call to get the updated categories after the user logs in, no?

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: I’d rather have this then a modal cause a modal to pop up. It’s just poor form.

sourcemash/frontend/static/js/routers/app_router.js
alexgerstein: Does it? What happens when you're not subscribed to any feeds? What categories appear in the sidenav after a browser refresh?

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: I don't think it's poor form to have one modal close and another to open. I think it's ugly to have <-- as if that gives any legitimate directions to the user when it's on the top right of the screen.

swglad: Let’s cut the toast then, but I don’t want to do a modal to modal. We’ll just hope the user notices.

sourcemash/api/users.py
swglad: Hmm. Idk. But it's working perfectly correctly. I can investigate.

sourcemash/frontend/static/js/routers/app_router.js
swglad: Just wiped my db but I'll run a seed and check later.

sourcemash/frontend/static/js/views/forgot_modal_view.js
swglad: It's an ugly and stumpy little form with one field, but okay.

sourcemash/frontend/static/js/views/register_modal_view.js
alexgerstein: And worse comes worst, they click on the form in the sidenav themselves.

sourcemash/api/categories.py
```diff
@@ -28,8 +28,11 @@ def get(self, id):
 
 class CategoryListAPI(Resource):
 
-    @login_required
     def get(self):
+
+        if not current_user.is_authenticated():
+          return {'categories': []}
+
```
swglad: ```current_user.subscribed``` (one line below) throws an error, so we need to check for authentication.

sourcemash/api/users.py
swglad: All works correctly. User comes to the frontend as undefined.

sourcemash/frontend/static/js/default.js
```diff
@@ -14,4 +16,12 @@ $(document).ajaxError(function (e, xhr, options) {
         fieldErrors.text(errors[errorField]);
     }
   }
-});
\ No newline at end of file
+
+  // Show Modal, if Not Authenticated
+  var errorStatus = xhr.status;
+  if (errorStatus == 401) {
```
alexgerstein: No point to save errorStatus as a var when we just use it once. Want to make it ```if (xhr.status == 401)```?

sourcemash/frontend/static/js/routers/app_router.js
```diff
@@ -13,11 +13,17 @@ Sourcemash.Routers.AppRouter = Backbone.Router.extend({
 
         self._user = new Sourcemash.Models.User();
         self._feeds = new Sourcemash.Collections.Feeds([], {allFeeds: true});
-        self._categories = new Sourcemash.Collections.Categories();
+        self._categories = new Sourcemash.Collections.Categories([]);
```
alexgerstein: Do we need the empty list now that we fixed the categories API call?

sourcemash/frontend/static/js/views/forgot_modal_view.js
```diff
@@ -0,0 +1,42 @@
+Sourcemash.Views.ForgotModalView = Backbone.View.extend({
+    template: JST['forgot-modal'],
+
+    events: {
+        'click .forgot-submit': 'recoverFromModal',
+        'submit #forgot-form': 'recoverFromModal',
+        'keyup #forgot-email': 'processKey',
```
alexgerstein: Do we need this?

sourcemash/frontend/static/js/views/register_modal_view.js
```diff
@@ -0,0 +1,53 @@
+Sourcemash.Views.RegisterModalView = Backbone.View.extend({
+    template: JST['register-modal'],
+
+    events: {
+        'click .register-submit': 'registerFromModal',
+        'submit #register': 'registerFromModal',
+        'keyup #register-password': 'processKey',
```
alexgerstein: Needed?

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -40,12 +42,21 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
+  },
+
   render: function() {
     this.close();
 
     activeTab = $(".tab .active").text().toLowerCase() || "categories";
 
-    var content = this.template({active: activeTab, current_user: this.user, feeds: this.feeds, categories: this.categories});
+    var content = this.template({active: activeTab, current_user: this.user, feeds: this.feeds,
+                                 categories: this.categories, isLoading: this.loading});
```
alexgerstein: Why do we need the isLoading here again?

sourcemash/frontend/static/js/views/register_modal_view.js
swglad: Yes. There is no “enter” event, you need to look for a keyup on 13.

sourcemash/frontend/static/scss/sourcemash.scss
```diff
@@ -187,7 +187,7 @@ a.button-collapse.top-nav {
 	}
 }
 
-.modal-close {
+.modal-action {
 	&:hover {
 		background-color: color("green", "base");
```
alexgerstein: Can we make it not green? Or at least a lighter one?

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: To fix the padding when the loading goes away, it was very easy to just add a check on isLoading and remove the padding (by applying class=“hide”) if it’s off.

sourcemash/frontend/static/scss/sourcemash.scss
swglad: Play with that as part of your UI update when you fix the color scheme.

sourcemash/frontend/templates/sidenav.tpl
```diff
@@ -11,7 +11,7 @@
       </div>
       <ul id="categories-list" class="collection col s12">
         <% if (current_user.get('email')) { %>
-          <div class="loading center-align"></div>
+          <div class="loading center-align <%= isLoading ? '' : 'hide'%>"></div>
```
alexgerstein: Did the css selector not work? ```.loading :not(.hide)``` should work. (Not the space).

sourcemash/frontend/static/scss/sourcemash.scss
alexgerstein: Will do.

sourcemash/frontend/static/js/views/sidenav_view.js
alexgerstein: It's nice when the other views are blind to the loading views. See my note below.

sourcemash/frontend/static/js/views/forgot_modal_view.js
swglad: Yes, see other answer.

sourcemash/frontend/templates/sidenav.tpl
swglad: Didn't try it with the space. I'll give that a shot.

sourcemash/frontend/static/js/default.js
swglad: The variable name makes the code a bit clearer, but I'm not emotionally committed if you want it gone.

alexgerstein: We could do xhr.error().status if you'd prefer since it's more readable.

swglad: Done.

sourcemash/frontend/static/js/routers/app_router.js
swglad: Good catch

sourcemash/frontend/templates/sidenav.tpl
swglad: Didn't work.

sourcemash/frontend/static/js/views/sidenav_view.js
swglad: Didn't work

sourcemash/frontend/static/js/routers/app_router.js
alexgerstein: You didn't change it though?

sourcemash/frontend/static/js/views/sidenav_view.js
alexgerstein: This will do it:

```css

.preloader-wrapper:not(.hide) {
  margin: 10px 0;
}

```

swglad: Are you sure / do you want to test it? Applying it to ```preloader-wrapper``` messed up other things (you were right to apply it to ```.loading```)

alexgerstein: Padding messed it up because that's how the preloader keeps itself centered. Margin won't cause issues because that pushes surrounding divs away, leaving the space we want. I just tried it locally and it worked.

PR #224, opened 2015-04-24
Try removing tpl.js.

alexgerstein:


swglad:
So everything still works without this there? Or did you not try it out yet?

If the latter, feel free to merge, play with out, check out the deploy, and then just comment that it was successful (or undo if not).

alexgerstein:
Haven’t tried it yet.

-- 
Alex Gerstein
gerstein.alex@gmail.com

On April 25, 2015 at 4:42:09 PM, Scott Gladstone (notifications@github.com) wrote:

So everything still works without this there? Or did you not try it out yet?

If the latter, feel free to merge, play with out, check out the deploy, and then just comment that it was successful (or undo if not).

PR #225, opened 2015-04-24
Fix category loading.

alexgerstein:


swglad:
Great category improvements on top of the loading fix!

PR #234, opened 2015-04-29
Production seed

### LINKED ISSUE #230: Seed Production with More Feeds.

swglad:
Closes #230.

### DIFF
--- a/feeds.json
+++ b/feeds.json
@@ -1,0 +1,321 @@
+{
+ "feeds": [
+    {
+      "tech": [
+        {
+          "title":"TechCrunch > Startups",
+          "url":"http://feeds.feedburner.com/techcrunch/startups?format=xml"
+        },
+        {
+          "title":"Engadget",
+          "url":"http://www.engadget.com/rss-full.xml"
+        },
+        {
+          "title":"Gizmodo",
+          "url":"http://feeds.gawker.com/gizmodo/full"
+        },
+        {
+          "title":"The Next Web",
+          "url":"http://thenextweb.com/feed/"
+        },
+        {
+          "title": "The Verge",
+          "url": "http://www.theverge.com/rss/full.xml"
+        },
+        {
+          "title": "Lifehacker",
+          "url": "http://feeds.gawker.com/lifehacker/vip"
+        },
+        {
+          "title": "ReadWrite",
+          "url": "http://www.readwriteweb.com/rss.xml"
+        },
+        {
+          "title": "Wired",
+          "url": "http://feeds.wired.com/wired/index"
+        },
+        {
+          "title": "Mashable",
+          "url": "http://feeds.mashable.com/Mashable"
+        },
+        {
+          "title": "Ars Technica",
+          "url": "http://feeds.arstechnica.com/arstechnica/index/"
+        },
+        {
+          "title": "Slashdot",
+          "url": "http://rss.slashdot.org/Slashdot/slashdot"
+        },
+        {
+          "title": "The Official Google Blog",
+          "url": "http://googleblog.blogspot.com/atom.xml"
+        },
+        {
+          "title": "Android Central",
+          "url": "http://www.androidcentral.com/feed"
+        },
+        {
+          "title": "MacRumors",
+          "url": "http://www.macrumors.com/macrumors.xml"
+        },
+        {
+          "title": "Boing Boing",
+          "url": "http://feeds.boingboing.net/boingboing/iBag"
+        },
+        {
+          "title": "ThinkGeek",
+          "url": "http://www.thinkgeek.com/thinkgeek.rss"
+        },
+        {
+          "title": "Forbes: Entrepreneurs",
+          "url": "http://www.forbes.com/entrepreneurs/index.xml"
+        },
+        {
+          "title": "A VC: Venture Capital and Technology",
+          "url": "http://feeds.feedburner.com/AVc"
+        },
+        {
+          "title": "VentureBeat",
+          "url": "http://feeds.feedburner.com/venturebeat"
+        },
+        {
+          "title": "Inc.com",
+          "url": "http://www.inc.com/rss.xml"
+        }
+      ]
+    },
+
+    {
+      "business": [
+        {
+          "title": "Seth Godin",
+          "url": "http://sethgodin.typepad.com/seths_blog/atom.xml"
+        },
+        {
+          "title": "Business Insider",
+          "url": "http://feeds2.feedburner.com/businessinsider"
+        },
+        {
+          "title": "Entrepreneuer: Latest Articles",
+          "url": "http://feeds.feedburner.com/entrepreneur/latest"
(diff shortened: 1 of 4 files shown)


worker/categorize.py
```diff
@@ -395,7 +395,7 @@ def _is_viable_candidate(self, phrase):
         words = phrase.split()
 
         # Ignore stop words
-        if words[0].lower() in STOP_WORDS:
+        if len(words) == 0 or words[0].lower() in STOP_WORDS:
```
swglad: Added this because I got this stack trace:

INFO:sqlalchemy.engine.base.Engine:(545,)
Traceback (most recent call last):
  File "manage.py", line 140, in <module>
    manager.run()
  File "/Users/scottgladstone/Sourcemash/flask/lib/python2.7/site-packages/flask_script/__init__.py", line 412, in run
    result = self.handle(sys.argv[0], sys.argv[1:])
  File "/Users/scottgladstone/Sourcemash/flask/lib/python2.7/site-packages/flask_script/__init__.py", line 383, in handle
    res = handle(*args, **config)
  File "/Users/scottgladstone/Sourcemash/flask/lib/python2.7/site-packages/flask_script/commands.py", line 216, in __call__
    return self.run(*args, **kwargs)
  File "manage.py", line 89, in feed_seed
    scrape_and_categorize_articles()
  File "/Users/scottgladstone/Sourcemash/worker/scraper.py", line 40, in scrape_and_categorize_articles
    categories = categorizer.categorize_item(item.title, text_only)
  File "/Users/scottgladstone/Sourcemash/worker/categorize.py", line 93, in categorize_item
    keyword_candidates = self._get_keyword_candidates(title, text)
  File "/Users/scottgladstone/Sourcemash/worker/categorize.py", line 116, in _get_keyword_candidates
    text_ngrams = self._get_valid_ngrams(text)
  File "/Users/scottgladstone/Sourcemash/worker/categorize.py", line 143, in _get_valid_ngrams
    if self._is_viable_candidate(ngram):
  File "/Users/scottgladstone/Sourcemash/worker/categorize.py", line 398, in _is_viable_candidate
    if words[0].lower() in STOP_WORDS:
IndexError: list index out of range

worker/scraper.py
```diff
@@ -99,15 +99,21 @@ def _store_items(feed):
...
-                    break
+            try:
+                img_tags = BeautifulSoup(requests.get(fp.feed.link).content).find_all('img')
+                for image_tag in img_tags:
+                    if "logo" in str(image_tag):
+                        feed.image_url = image_tag.get('src') or image_tag.get('href')
+                        db.session.commit()
+                        break
+            except:
+                pass # eventually, default image
```
swglad: Added this try block b/c some RSS feeds apparently don't include a ```<link>``` entry in their XML. The comment is because we actually do want default images there eventually.

feeds.json
```diff
@@ -0,0 +1,321 @@
...
+          "url": "http://www.economist.com/rss/the_world_this_week_rss.xml"
+        },
+        {
+          "title": "Bloomberg",
+          "url": "http://www.newslookup.com/rss/business/bloomberg.rss"
+        }
+      ]
+    }
+  ]
+}
```
swglad: 67 feeds, if you were curious.

Taken from highly subscribed feeds (80k+) on feedly, using their categories as the topics (e.g. tech, gaming, finance)

worker/scraper.py
```diff
@@ -99,15 +99,21 @@ def _store_items(feed):
...
-                    break
+            try:
+                img_tags = BeautifulSoup(requests.get(fp.feed.link).content).find_all('img')
+                for image_tag in img_tags:
+                    if "logo" in str(image_tag):
+                        feed.image_url = image_tag.get('src') or image_tag.get('href')
+                        db.session.commit()
+                        break
+            except:
+                pass
```
swglad: Added this try block b/c some RSS feeds apparently don't include a ```<link>``` entry in their XML.

feeds.json
alexgerstein: did you type this yourself?

worker/categorize.py
alexgerstein: I would put a check for None/empty at the beginning of the function that returns false instead.

worker/scraper.py
alexgerstein: Instead of try block (since you do nothing on the catch, want to just do ```if fp.feed.link:``` instead?

manage.py
```diff
@@ -68,6 +68,25 @@ def worker(kill=False):
...
 
+@manager.command
+def feed_seed():
+    """Add seed feeds from JSON file"""
+    with open(FEED_DATA_FILE) as data_file:
+        data = json.load(data_file)
+        feeds_by_topic = data["feeds"]
+
+        for topic in feeds_by_topic:
+            for feed_json in topic[topic.keys()[0]]:
```
alexgerstein: Can do ```for feed_json in topic.values()[0]``` instead.

feeds.json
swglad: Picked the links myself, but wrote a script to convert it to json format.

alexgerstein: Very nice.

worker/categorize.py
swglad: There's already one there. The problem is after the split.

worker/scraper.py
swglad: Actually I raised an issue about this, but the except will be to set a default image (eg. Logo) as the image

worker/categorize.py
alexgerstein: So what input is it failing on?

worker/scraper.py
alexgerstein: Ah gotcha. Looks good then.

worker/categorize.py
swglad: I don't know. It failed after running for 30 mins and we don't have logging before this point to see what the ```phrase``` being split actually is.

alexgerstein: Ok. So if len(words) was 0. This means phrase.split() returned []. This can happen when phrase is " ". However, that wouldn't make it past the < 2 check at the beginning of the function. What about "  " (two spaces)? That would also return []. Same with any number of spaces. Maybe we should call strip() to start the function? Then the len(phrase) < 2 will return false for us?

swglad: I like that. Done.

worker/categorize.py
```diff
@@ -381,6 +381,7 @@ def _get_best_keywords(self, communities, original_keywords):
     def _is_viable_candidate(self, phrase):
 
         # Ignore 1-character phrases
+        phrase = phrase.strip()
```
alexgerstein: Actually, this isn't good because words with spaces will be stored into the category db with their spaces. Let's take this strip and move it before _is_viable_candidate is even called to make our categories cleaner. Wikipedia would also have issues with words with spaces at either end.

PR #237, opened 2015-04-30
Show limited categories to logged in user. All to not logged in users.

### LINKED ISSUE #236: Categories (N > 1) on item cards
To match categories present in tab menu. Quick fix as part of UI update.

alexgerstein:
Closes #236.

### DIFF
--- a/sourcemash/frontend/static/js/views/item_card_view.js
+++ b/sourcemash/frontend/static/js/views/item_card_view.js
@@ -95,6 +95,15 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
     },
 
     render: function() {
-        this.$el.html(this.template({ item: this.model, user: this.user }));
+        var shownCategories;
+        if (this.user.get('email')) {
+            shownCategories = this.model.categories.filter(function(category) {
+              return category.get("item_count") > 1;
+            });
+        } else {
+            shownCategories = this.model.categories.first(5)
+        }
+
+        this.$el.html(this.template({ item: this.model, categories: shownCategories, user: this.user }));
     }
 });
--- a/sourcemash/frontend/static/js/views/item_card_view.js
+++ b/sourcemash/frontend/static/js/views/item_card_view.js
@@ -3,6 +3,7 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
 
     initialize: function(options) {
         _.bindAll(this, "savedToast", "voted");
+        this.user = options.user;
         this.listenTo(this.model, 'change', this.render);
         this.render();
     },
@@ -94,6 +95,6 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
     },
 
     render: function() {
-        this.$el.html(this.template({ item: this.model }));
+        this.$el.html(this.template({ item: this.model, user: this.user }));
     }
 });
--- a/sourcemash/frontend/static/js/views/item_card_view.js
+++ b/sourcemash/frontend/static/js/views/item_card_view.js
@@ -101,9 +101,9 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
               return category.get("item_count") > 1;
             });
         } else {
-            shownCategories = this.model.categories.first(5)
+            shownCategories = this.model.categories.first(Sourcemash.Views.ItemCardView.MAX_CATEGORIES_SHOWN)
         }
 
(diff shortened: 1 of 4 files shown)


sourcemash/frontend/templates/item-card.tpl
```diff
@@ -17,7 +17,9 @@
       <div class="card-action valign-wrapper row">
         <div class="category-list flow-text col s8">
           <% for (var i = 0; i < item.categories.length; i++) { %>
-            <a href="/#categories/<%=item.categories.at(i).get('id')%>"><%=item.categories.at(i).get('name')%></a>
+            <% if (item.categories.at(i).get('item_count') > 1 || !user.get('email')) { %>
```
swglad: When we have too many categories, the cards don't look that great and it won't mean much to a not-logged-in user anyway. Can we cap the number of categories at 5 or 6 for a not logged in user (aka add an if statement to break after ```i``` reaches a certain point?)?

alexgerstein: Fixed.

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -94,6 +95,15 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
     },
 
     render: function() {
-        this.$el.html(this.template({ item: this.model }));
+        var shownCategories;
+        if (this.user.get('email')) {
+            shownCategories = this.model.categories.filter(function(category) {
+              return category.get("item_count") > 1;
```
swglad: Why the return?

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -94,6 +95,15 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
+            });
+        } else {
+            shownCategories = this.model.categories.first(5)
+        }
+
+        this.$el.html(this.template({ item: this.model, categories: shownCategories, user: this.user }));
```
swglad: Does user need to be passed through anymore?

sourcemash/frontend/static/js/views/item_card_view.js
swglad: Wouldn't that stop the HTML from being filled in line 107 below?

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -94,6 +95,15 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
 
     render: function() {
-        this.$el.html(this.template({ item: this.model }));
+        var shownCategories;
+        if (this.user.get('email')) {
+            shownCategories = this.model.categories.filter(function(category) {
+              return category.get("item_count") > 1;
+            });
+        } else {
+            shownCategories = this.model.categories.first(5)
```
swglad: Back out this magic number to the view's initialize as MAX_CATEGORIES?

sourcemash/frontend/static/js/views/item_card_view.js
alexgerstein: Nope. Fixed.

sourcemash/frontend/static/js/views/item_card_view.js
```diff
@@ -94,6 +95,15 @@ Sourcemash.Views.ItemCardView = Backbone.View.extend({
...
+              return category.get("item_count") > 1;
+            });
+        } else {
+            shownCategories = this.model.categories.first(Sourcemash.Views.ItemCardView.MAX_CATEGORIES_SHOWN)
+        }
+
+        this.$el.html(this.template({ item: this.model, categories: shownCategories }));
     }
-});
+}, {MAX_CATEGORIES_SHOWN: 5});
```
swglad: To confirm, this is being passed as a second dictionary to the entire view, which allows you to access it via the line ```Sourcemash.Views.ItemCardView.MAX_CATEGORIES_SHOWN```?

alexgerstein: Right. The second dictionary just allows you to pass whatever values you want to be added to the view class. It would be the same as if I typed ```Sourcemash.Views.ItemCardView.MAX_CATEGORIES = 5```. It adds the property onto all instances of the view so that I can access it like above.

swglad: Awesome.

PR #241, opened 2015-05-03
Make categories always present.

### LINKED ISSUE #207: Upvote/save INSIDE the item card
With the inside having a cancel/close card button instead of the list of categories:

Cancel ... ... ... ... Saved Vote

alexgerstein:
Closes #207.

### DIFF
--- a/sourcemash/frontend/templates/item-card.tpl
+++ b/sourcemash/frontend/templates/item-card.tpl
@@ -13,25 +13,6 @@
         </div>
       <div class="col s1"><i class="mdi-navigation-more-vert activator right"></i></div>
       </div>
-
-      <div class="card-action valign-wrapper row">
-        <div class="category-list flow-text col s8">
-          <% _.each(categories, function(category) { %>
-            <a href="/#categories/<%=category.get('id')%>"><%=category.get('name')%></a>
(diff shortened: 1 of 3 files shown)


swglad:
Cards are getting "knocked out of place" in their row and leaving empty spaces. Is there a way to either make both cards in each row the same length (like before) or have the cards be offset from one another and not always line up (like notey: http://www.notey.com/blogs/outfit)?

screen shot 2015-05-03 at 6 00 32 pm

sourcemash/frontend/templates/item-card.tpl
```diff
@@ -43,4 +24,22 @@
       <p class="flow-text summary"><%= item.get('summary')%></p>
     </div>
   </div>
+
+  <div class="card-action z-depth-1 valign-wrapper row">
+    <div class="category-list flow-text col s8">
+      <% _.each(categories, function(category) { %>
```
swglad: Only categories N > 1

alexgerstein: Fixed.

alexgerstein:
The offset should be fixed now that the categories > 1 should be resolved.

PR #251, opened 2015-05-05
dendogram bug fix(?)

### LINKED ISSUE #240: Dendogram size error
Still present in categorization. It appears to be occurring with feeds that are really comics/videos.

swglad:
Closes #240.

### DIFF
--- a/worker/categorize.py
+++ b/worker/categorize.py
@@ -104,7 +104,8 @@ class Categorizer:
 
         if len(assigned_articles) > 2:
             semantic_graph = self._build_semantic_graph(assigned_articles)
-            clustering = community.best_partition(semantic_graph)
+            if semantic_graph.graph:
+                clustering = community.best_partition(semantic_graph)
 
         # Use keywords from best clusters
         selected_categories = self._get_best_keywords(clustering, keyword_candidates)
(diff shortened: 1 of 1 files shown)


worker/categorize.py
```diff
@@ -101,9 +101,8 @@ def categorize_item(self, title, text):
 
         # Build a graph and identify keyword clusters
         clustering = None
-
-        if len(assigned_articles) > 2:
-            semantic_graph = self._build_semantic_graph(assigned_articles)
+        semantic_graph = self._build_semantic_graph(assigned_articles)
+        if semantic_graph.edges():
```
swglad: Looks great. Did you try running ```python manage.py feed_seed``` and letting the categorization part run for a little while and see if anything breaks?

alexgerstein: I didn't do feed_seed to test, but I wrote a test that was returning the dendogram error and added this code to get it to pass.

I removed the test because I didn't feel it was necessary to keep once we fixed it.

swglad: Can you run feed seed for an hour and just see if you get any bugs?

alexgerstein: Sure.

PR #255, opened 2015-05-07
Add item-card specification.

### LINKED ISSUE #252: Fix item card formatting
Because we're getting empty spaces.

3 column solution: 2 for medium/large screen and 1 for small screen that has full list time order.

alexgerstein:
Closes #252.

### DIFF
--- a/sourcemash/frontend/static/scss/sourcemash.scss
+++ b/sourcemash/frontend/static/scss/sourcemash.scss
@@ -379,7 +379,6 @@ a.button-collapse.top-nav {
     &.row {
       margin: 0 auto;
       margin-bottom: 10px;
-      min-height: 90px;
     }
 
     .category-list {
--- a/sourcemash/frontend/static/scss/sourcemash.scss
+++ b/sourcemash/frontend/static/scss/sourcemash.scss
@@ -367,47 +367,50 @@ a.button-collapse.top-nav {
     }
 }
 
-.card-action {
-  @extend .grey-text;
+.item-card {
+  .card-action {
+    @extend .grey-text;
 
-  padding: 10px;
-  padding-left: 0;
-  border: 1px solid rgba(160,160,160,.2);
-  width: 100%;
+    padding: 10px;
+    padding-left: 0;
+    border: 1px solid rgba(160,160,160,.2);
+    width: 100%;
 
-  &.row {
-    margin: 0 auto;
-    margin-bottom: 10px;
-    min-height: 90px;
-  }
+    &.row {
+      margin: 0 auto;
+      margin-bottom: 10px;
+      min-height: 90px;
+    }
 
-  .category-list {
-    a {
-      @extend .grey-text;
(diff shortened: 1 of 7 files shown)


swglad:
This looks like it may resolve (or help to resolve) #257, but doesn't do what #252 specifies.

swglad:
@alexgerstein Did you see the above comment?

alexgerstein:
Ah. I'm on it.

swglad:
Also probably goes without saying, but can you run feed-seed and make sure the feed cards look nice given the set of feeds we have loaded? If we fix our design to accommodate this set of feeds, that's not a bad thing

alexgerstein:
How else was I going to make sure the two-col solution was working!?!

swglad:
Talking about two separate but related issues: two vs. three columns is one, but style of feed cards vs. item cards is another. Just making sure we maximize both!

alexgerstein:
@swglad It's ready.

sourcemash/frontend/templates/category.tpl
```diff
@@ -7,11 +7,25 @@
 
   <div class="loading center-align"></div>
 
-  <ul id="items" class="list-group row">
-    <% items.forEach(function(item) { %>
-        <div id="item-<%= item.get('id') %>" class="item-card"></div>
-    <% }); %>
-  </ul>
+  <div id="items" class="list-group row">
+    <div class="col m6 hide-on-small-and-down">
```
swglad: What's smaller than small? (not a riddle)

sourcemash/frontend/templates/category.tpl
```diff
@@ -7,11 +7,25 @@
...
+      <% for (i = 0; i < items.length; i = i + 2) { %>
```
swglad: Any reason for this type of for loop as opposed to the forEach we usually use?

swglad: Ah I see how you split the two columns. I was imagining one loop with two div's inside it that alternated indices with if statements. Looks good!

sourcemash/frontend/static/scss/sourcemash.scss
```diff
@@ -327,48 +322,50 @@ a.button-collapse.top-nav {
     }
 }
 
-.card-action {
-  @extend .grey-text;
+.item-card {
```
swglad: I noticed you didn't make any style changes on feed-cards. Everything looks good on that end (e.g. feed descriptions don't run too long and take over the card, subscribe buttons don't get pushed by long titles, etc)?

sourcemash/frontend/templates/category.tpl
alexgerstein: No, but might be a rap lyric...

Small is just 600px, small-and-down sets max-width of 600px.

swglad:
Why did my merging the MixPanel branch just merge this one?

PR #256, opened 2015-05-07
Get all categories when not logged in.

### LINKED ISSUE #239: Non-logged in users click on categories
Should it (1) show all items in that category, or (2) prompt them with a modal to log in? 

This is if they click on a category on an item card. Currently takes you to an empty category page (b/c api/categories returns []) but I think it should probably show all items.

alexgerstein:
Closes #239.

### DIFF
--- a/sourcemash/frontend/static/js/routers/app_router.js
+++ b/sourcemash/frontend/static/js/routers/app_router.js
@@ -65,7 +65,7 @@ Sourcemash.Routers.AppRouter = Backbone.Router.extend({
         var category = new Sourcemash.Models.Category({ id: id });
         category = this._categories.add(category);
 
-        var categoryItems = new Sourcemash.Collections.Items([], {category: category});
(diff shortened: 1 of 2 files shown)


sourcemash/frontend/static/js/routers/app_router.js
```diff
@@ -65,7 +65,7 @@ Sourcemash.Routers.AppRouter = Backbone.Router.extend({
         var category = new Sourcemash.Models.Category({ id: id });
         category = this._categories.add(category);
 
-        var categoryItems = new Sourcemash.Collections.Items([], {category: category});
+        var categoryItems = new Sourcemash.Collections.Items([], {category: category, allItems: this._user.get('email') ? false : true});
```
swglad: Awesome. Out of curiosity, how did you test this?

alexgerstein: I went to a categories page once when logged in and not subscribed to anything, and once when not. I made the change so that only one item showed up when I was logged in, and three when not.

PR #258, opened 2015-05-07
Browse feed updates

### LINKED ISSUE #249: Browse feeds page split into topics
"Mash" topic aka pending/waiting to be categorized into another pre-existing topic

swglad:
Resolves #249.

Pull and check out the browse page!


sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -32,24 +33,33 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
                                     "Source": 'browse' });
   },
 
+  preserveURL: function(e){
+    e.preventDefault();
+    window.location.href = e.target.href;
+    window.location.replace('/#browse');
+  },
```
swglad: 1. Stops the href from HTML from going.
2. Extracts the href from the HTML and goes there.
3. Replaces the address bar with "/#browse"

manage.py
```diff
@@ -80,8 +80,13 @@ def feed_seed():
 
                 feed = Feed(title=feed_json["title"],
                             url=feed_json["url"],
+                            topic=topic.keys()[0],
```
alexgerstein: What is this if it's just topic instead of topic.keys()[0]?

migrations/versions/293250007b39_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message
```
alexgerstein: Did you run this db upgrade and db downgrade locally to make sure it works?

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -34,22 +34,28 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
 
   render: function() {
     // Render parent view
-    this.$el.html(this.template({ models: this.collection.models }));
+    this.$el.html(this.template({ models: this.feedTopicViews }));
```
alexgerstein: Won't this be empty on the first render? Maybe we should just pass through the topics array and rename models to topics.

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -34,22 +34,28 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
...
-    var feedCards = [];
-    this.collection.models.forEach(function(feed) {
-        var feedCardView = new Sourcemash.Views.FeedCardView({el: "#feed-card-" + feed.get('id'),
-                                                              model: feed});
-        feedCards.push(feedCardView)
+    feeds = this.collection;
+    var topicViews = [];
+    Sourcemash.Views.BrowseView.FEED_TOPICS.forEach(function(topic) {
+      if (feeds.length > 1) {
+        var topicCards = feeds.where({topic: topic});
```
alexgerstein: Aren't these feed cards? Not topic cards?

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -34,22 +34,28 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
...
 
-    // Render item cards
-    var feedCards = [];
-    this.collection.models.forEach(function(feed) {
-        var feedCardView = new Sourcemash.Views.FeedCardView({el: "#feed-card-" + feed.get('id'),
-                                                              model: feed});
-        feedCards.push(feedCardView)
+    feeds = this.collection;
+    var topicViews = [];
+    Sourcemash.Views.BrowseView.FEED_TOPICS.forEach(function(topic) {
```
alexgerstein: Sweet fix to not need the FEED_TOPICS array: http://stackoverflow.com/questions/9954597/how-to-use-groupby-in-backbone-js-to-group-collections

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -34,22 +34,28 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
...
     this.$(".loading").html(this.loadingView.render().el);
 
-    // Render item cards
-    var feedCards = [];
-    this.collection.models.forEach(function(feed) {
-        var feedCardView = new Sourcemash.Views.FeedCardView({el: "#feed-card-" + feed.get('id'),
-                                                              model: feed});
-        feedCards.push(feedCardView)
+    feeds = this.collection;
+    var topicViews = [];
```
alexgerstein: Why do we need this when we already have the this.feedTopicViews?

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -71,10 +77,10 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
   },
 
   close: function() {
-    _.each(this.feedCardViews, function(feedCardView) {
-        feedCardView.close();
-        feedCardView.remove();
-        feedCardView.unbind();
+    _.each(this.feedTopicViews, function(feedTopicView) {
```
alexgerstein: You should call this.close() in the render function so we don't have duplicate views lying around.

sourcemash/frontend/static/js/views/feed_topic_view.js
```diff
@@ -0,0 +1,38 @@
+Sourcemash.Views.FeedTopicView = Backbone.View.extend({
+  template: JST['feed-topic'],
+
+  initialize: function(options) {
+    this.collection = options.collection;
```
alexgerstein: This line happens auto-magically with backbone views (same for this.model).

sourcemash/frontend/static/js/views/feed_topic_view.js
```diff
@@ -0,0 +1,38 @@
...
+                                                              model: feed});
+        feedCards.push(feedCardView);
+    });
+
+    this.feedCardViews = feedCards;
+    return this;
+  },
+
+  close: function() {
+    _.each(this.feedCardViews, function(feedCardView) {
```
alexgerstein: Call this.close in the render function.

sourcemash/frontend/templates/feed-topic.tpl
```diff
@@ -0,0 +1,16 @@
+<div id="<%= topic %>" class="section scrollspy">
```
alexgerstein: Move this line to browse.tpl.

sourcemash/frontend/templates/browse.tpl
```diff
@@ -11,10 +11,31 @@
...
+      <% models.forEach(function(model) { %>
+        <div id="feed-topic-<%= model.topic %>" class="feed-topic"></div>
+      <% }); %>
+    </div>
+
+    <div class="col hide-on-small-only m2 l2">
+      <div class="tabs-wrapper">
+        <h5 class="contents-header grey-text text-darken-2">Topics</h5>
+        <ul class="section table-of-contents">
+          <li><a href="#Technology">Technology</a></li>
```
alexgerstein: Do we want to display these automatically using the same for loop for the divs?

manage.py
swglad: topic is really topic_dict. I can rename if confusing.

migrations/versions/293250007b39_.py
swglad: Yup, both work!

sourcemash/frontend/static/js/views/feed_topic_view.js
swglad: Cool. Easy change.

sourcemash/frontend/static/js/views/browse_view.js
swglad: So the advantage of the array is that we can control the order. I think we should keep it, and I'll pass through that to the template to set up the table of contents + feed topic headers.

sourcemash/frontend/static/js/views/browse_view.js
swglad: They're feed cards within a topic. I rename them feedCards.

sourcemash/frontend/static/js/views/browse_view.js
swglad: Done.

sourcemash/frontend/static/js/views/feed_topic_view.js
swglad: Done.

sourcemash/frontend/templates/browse.tpl
swglad: Done.

sourcemash/frontend/templates/feed-topic.tpl
swglad: I think scrollspy should be attached to the same ID as the href from the table of contents points to, so I'd prefer to leave it here.

sourcemash/frontend/static/js/views/browse_view.js
swglad: Much better idea. Will do.

sourcemash/frontend/static/js/views/browse_view.js
swglad: If we add this.close to the render, will we need to maintain this.feedTopicViews at all?

swglad: Nevermind!

swglad: So nothing renders if you try to just use this.feedTopicViews. I used the same structure here that you did when writing the initial browse_view.js which is now in feed_topic_view.js, so is this something that you encountered and dealt with in your initial writing?

alexgerstein: We still need this.feedTopicViews because the view needs to know which subviews to close.

You were probably experiencing a scoping issue since the forEach loop has a nested function. It didn't know what this.feedTopicViews was referring to the "this" outside the nested function. What's nice about the forEach function is that you can pass an additional parameter for context. I'll mark where to put the "this" below.

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -33,23 +33,31 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
...
+    feeds = this.collection;
+    var topicViews = [];
+    Sourcemash.Views.BrowseView.FEED_TOPICS.forEach(function(topic) {
+      if (feeds.length > 1) {
+        var feedCards = feeds.where({topic: topic});
+        var feedTopicView = new Sourcemash.Views.FeedTopicView({collection: feedCards, topic: topic,
+                                                                el: "#feed-topic-" + topic});
+        topicViews.push(feedTopicView);
+      };
     });
```
alexgerstein: ```}, this);```

sourcemash/frontend/static/js/views/browse_view.js
alexgerstein: Sounds good.

manage.py
```diff
@@ -75,13 +75,18 @@ def feed_seed():
...
+        for topic_json in feeds_by_topic:
+            for feed_json in topic_json.values()[0]:
 
                 feed = Feed(title=feed_json["title"],
                             url=feed_json["url"],
+                            topic=topic_json.keys()[0],
                             last_updated = datetime.min)
 
+                # Don't re-add existing feed
+                if Feed.query.filter_by(url=feed.url).first():
```
alexgerstein: Let's not construct the feed object in the first place if it's a duplicate. So ```if not Feed.query.filter_by(url=feed_json['url']).first():```

sourcemash/frontend/static/js/views/browse_view.js
```diff
@@ -33,23 +33,29 @@ Sourcemash.Views.BrowseView = Backbone.View.extend({
...
-    // Render item cards
-    var feedCards = [];
-    this.collection.models.forEach(function(feed) {
-        var feedCardView = new Sourcemash.Views.FeedCardView({el: "#feed-card-" + feed.get('id'),
-                                                              model: feed});
-        feedCards.push(feedCardView)
-    });
+    feeds = this.collection;
+    Sourcemash.Views.BrowseView.FEED_TOPICS.forEach(function(topic) {
+      if (feeds.length > 1) {
```
alexgerstein: What's this check used for?

swglad: The render sometimes gets passed through twice, once with feeds = [] and another time with feeds full of stuff. This makes sure you don't try to start making feeds and views when there isn't anything to make.

manage.py
```diff
@@ -75,15 +75,18 @@ def feed_seed():
...
 
                 feed = Feed(title=feed_json["title"],
                             url=feed_json["url"],
+                            topic=topic_json.keys()[0],
                             last_updated = datetime.min)
 
-                db.session.add(feed)
-                db.session.commit()
+                # Don't re-add existing feed
+                if not Feed.query.filter_by(url=feed.url).first():
```
alexgerstein: I meant move this to above ```feed = Feed```....

sourcemash/frontend/static/js/views/browse_view.js
alexgerstein: It should still work without this line though, right?

PR #289, opened 2015-05-13
Admin Page

### LINKED ISSUE #120: Admins: assign role, privileged API calls
Scott Gladstone
Should there be a delete to actually remove a feed from the database? It's not something that user's should be able to do, but maybe admins? For example, if an admin wanted to remove an "offensive" feed?

It would have both login required and role required decorators.

Alex Gerstein  
alexgerstein added a note 26 minutes ago
That's a good point. We might want that as an option in the future. Unsubscribe is the more common call though, so I think it makes sense to leave it as the DELETE to FeedList for now.

If we want to add an admin delete in the future, we can add a feeds/id/delete URL with role protection (admins-only) on it down the line.

Scott Gladstone  
Unsubscribe is currently a delete in FeedAPI. We'd put admin delete in FeedListAPI, and perhaps have the admin delete button send the name of the feed in a form?

Scott Gladstone  
swglad added a note 22 minutes ago
Would you want to add it now? I'd only encourage it because we're making enormous API changes here so it makes sense as part of this PR more than as a future issue.

Alex Gerstein   ]
alexgerstein added a note 12 minutes ago
You don't want to put a DELETE on a list, since that route doesn't specify which feed you're trying to delete. Since we don't have any other admin commands yet--and deleting inappropriate feeds is more of an enhancement--I think we can take time to consider how we handle our admin API and create an "admin permissions" task for the future.

Also remember we don't even have ways of assigning people the admin role yet.

alexgerstein:
Closes #120.

### DIFF
--- a/sourcemash/admin.py
+++ b/sourcemash/admin.py
@@ -1,0 +1,22 @@
+from flask.ext.admin import Admin
+from flask.ext.admin.contrib.sqla import ModelView
+from flask.ext.login import current_user
+from sourcemash.database import db
+from sourcemash.models import Feed
+
+admin = Admin()
+
+
+class FeedView(ModelView):
+    can_create = False
+    can_delete = False
+
+    def is_accessible(self):
+        return current_user.is_authenticated() and \
+               current_user.email == "admin@sourcemash.com"
+
+    column_display_pk = True
+    form_columns = ['title', 'topic', 'public', 'image_url']
+
+
+admin.add_view(FeedView(Feed, db.session))
--- a/sourcemash/admin.py
+++ b/sourcemash/admin.py
@@ -1,3 +1,4 @@
+from flask import redirect, url_for
 from flask.ext.admin import Admin
 from flask.ext.admin.contrib.sqla import ModelView
 from flask.ext.login import current_user
@@ -14,6 +15,10 @@ class FeedView(ModelView):
     def is_accessible(self):
         return current_user.is_authenticated() and \
                current_user.email == "admin@sourcemash.com"
+
+    def _handle_view(self, name, **kwargs):
+            if not self.is_accessible():
+                return redirect(url_for('security.login'))
 
     column_display_pk = True
     form_columns = ['title', 'topic', 'public', 'image_url']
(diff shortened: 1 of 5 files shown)


sourcemash/admin.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def _handle_view(self, name, **kwargs):
+            if not self.is_accessible():
+                return redirect(url_for('security.login'))
+
+    column_display_pk = True
+    form_columns = ['title', 'topic', 'public', 'image_url']
+
+
+admin.add_view(FeedView(Feed, db.session))
```
swglad: Is the admin page at ```/admin```?

alexgerstein: Yessir

Alex Gerstein
gerstein.alex@gmail.com




On Wed, May 13, 2015 at 9:56 AM -0700, "Scott Gladstone" <notifications@github.com> wrote:












In sourcemash/admin.py:
> +    can_create = False
> +    can_delete = False
> +
> +    def is_accessible(self):
> +        return current_user.is_authenticated() and \
> +               current_user.email == "admin@sourcemash.com"
> +
> +    def _handle_view(self, name, **kwargs):
> +            if not self.is_accessible():
> +                return redirect(url_for('security.login'))
> +
> +    column_display_pk = True
> +    form_columns = ['title', 'topic', 'public', 'image_url']
> +
> +
> +admin.add_view(FeedView(Feed, db.session))



Is the admin page at /admin?

swglad: Can you add a link to it on the profile page if user.get(‘email’) == “admin@sourcemash.com”?

alexgerstein: Should I just put the link in the sidenav?

swglad: Sounds good. As long as it’s only Admin who can see it, I’m game for that. Gotta find a good icon for it then!

sourcemash/frontend/templates/sidenav.tpl
```diff
@@ -65,6 +65,9 @@
       <a href="/#browse"><li class="collection-item"><i class="mdi-action-view-module"></i>Browse</li></a>
       <a href="/#saved"><li class="collection-item"><i class="mdi-action-bookmark"></i>Saved</li></a>
       <a href="/#profile"><li class="collection-item"><i class="mdi-social-person"></i>Profile</li></a>
+      <% if (current_user.get('email')) { %>
```
swglad: ```== "admin@sourcemash.com"```?

PR #291, opened 2015-05-13
Split categorize and scrape.

### LINKED ISSUE #196: Split scrape and categorization tasks.
We should make the queue divide up the scraping tasks so more urgent tasks are dealt with sooner.

alexgerstein:
Closes #196.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -14,7 +14,8 @@ from sourcemash.database import db
 from sourcemash.models import User, Feed, Item, Category, UserItem
 
 from worker import create_worker
-from worker.scraper import scrape_and_categorize_articles
+from worker.categorize import Categorizer
+from worker.scraper import scrape_feed_articles, categorize_feed_articles
 
 from datetime import datetime
 import json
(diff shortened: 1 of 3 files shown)


manage.py
```diff
@@ -50,17 +52,25 @@ def test(all=False):
...
 
-    else:
-        q = Queue('default', connection=conn)
-        while True:
-            job = q.enqueue_call(func=scrape_and_categorize_articles, timeout=1800)
-            time.sleep(THIRTY_MINUTES)
+        for feed in Feed.query.all():
+            q.enqueue_call(func=categorize_feed_articles,
+                           args=(feed, categorizer,), timeout=1800)
+        time.sleep(THIRTY_MINUTES)
```
swglad: Delete this?

manage.py
```diff
@@ -14,7 +14,8 @@
 from sourcemash.models import User, Feed, Item, Category, UserItem
 
 from worker import create_worker
-from worker.scraper import scrape_and_categorize_articles
```
swglad: ```scrape_and_categorize_articles``` is still present in other commands in this file!

alexgerstein: That was careless. Let me fix that.

manage.py
alexgerstein: I think we should keep it. Because we're just enqueuing the tasks, it happens in seconds. If we get rid of the sleep, a categorize_feed_articles task will get added to the queue every second. This will lead to a big backlog of tasks that I don't think we want.

manage.py
alexgerstein: Should be fixed now.

PR #292, opened 2015-05-13
Remove nested categories.

### LINKED ISSUE #272: Remove nested categories

alexgerstein:
Closes #272.

### DIFF
--- a/worker/categorize.py
+++ b/worker/categorize.py
@@ -410,6 +410,11 @@ class Categorizer:
         return True
 
     def _is_nested(self, keyword, other_keywords):
+        '''
+        Check whether a keyword is already covered by a more specific one.
+        For example, "Google" is nested in "Google Maps." But "Google Maps" is
+        not nested in "Google."
+        '''
         for other_keyword in other_keywords:
             if keyword != other_keyword and keyword in other_keyword:
                 return True
--- a/worker/categorize.py
+++ b/worker/categorize.py
@@ -375,7 +375,8 @@ class Categorizer:
             best_original_keywords = keyword_counts.most_common(2)
             best_keywords.update([keyword[0] for keyword in best_original_keywords])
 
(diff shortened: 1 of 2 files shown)


worker/categorize.py
```diff
@@ -403,11 +404,18 @@ def _is_viable_candidate(self, phrase):
...
 
             titled_words = map(lambda x: x.istitle(), words)
-            if titled_words not in [[True, False, True], [True, True,True]]:
+            if titled_words not in [[True, False, True], [True, True, True]]:
                 return False
 
         return True
 
+    def _is_nested(self, keyword, other_keywords):
+        for other_keyword in other_keywords:
```
swglad: Wouldn't this make it possible to accidentally remove both "Google" and "Google Maps"? Does this run for every keyword against the list of all other keywords (e.g. Will it run with both "Google" and "Google Maps" as the keyword?).

worker/categorize.py
```diff
@@ -403,11 +404,18 @@ def _is_viable_candidate(self, phrase):
...
+            if keyword != other_keyword and keyword in other_keyword:
```
swglad: Should we favor longer keywords over shorter, or build in anything like that? Should "Google Maps" be chosen over "Google"? What happens now? Is it just luck of which runs first?

alexgerstein: Both "Google Maps" and "Google" will run as the keyword, but we don't need to worry about both being removed because ```"Google Maps" in "Google"``` returns ```false```. This approach will always take the longer option. For example, the Ebola/West Africa test now only returns "Ebola Virus in West Africa" and "Aylward" and "Sierra Leone."

swglad: Great. Can you make a comment with that?

alexgerstein: That's just how searching works. I don't know if it needs a comment for that particular line, but I can document the function.

PR #293, opened 2015-05-13
Read/Unread count update

### LINKED ISSUE #245: Remove unread/read count, just "new"

### LINKED ISSUE #278: Change category sort to 'title'

swglad:
Resolves #245. Closes #278.


sourcemash/api/categories.py
```diff
@@ -1,11 +1,20 @@
...
-from sourcemash.models import Item, Category, UserItem
+from sourcemash.models import Item, Category, UserItem, UserCategory
 
 from sqlalchemy import func
 
+class isUnread(fields.Raw):
+    def output(self, key, category):
+        try:
+            unread = UserCategory.query.filter_by(user=current_user, category_id=category.id).one().unread
+        except:
```
alexgerstein: Let's specify this except, like seen here: http://stackoverflow.com/questions/18110033/getting-first-row-from-sqlalchemy

sourcemash/api/categories.py
```diff
@@ -13,18 +22,46 @@
...
 
 category_status_fields = {
-  'item_count': fields.Integer,
-  'unread_count': fields.Integer
+  'unread': isUnread
 }
 category_status_fields = dict(category_fields, **category_status_fields)
 
+category_status_count_fields = {
+  'item_count': fields.Integer,
```
alexgerstein: Do we still want to bother with item count at all?

swglad: We do it for feeds so is it weird not to do for categories? That being said, I’m not too attached and wouldn’t have any problem cutting it.

sourcemash/api/categories.py
```diff
@@ -44,24 +81,16 @@ def get(self):
                              .all()
 
         for category, count in categories:
-            read_item_count = UserItem.query \
```
alexgerstein: It's beautiful seeing this disappear...

sourcemash/api/feeds.py
```diff
@@ -27,21 +27,17 @@ def output(self, key, feed):
...
-class getUnreadCount(fields.Raw):
+class isUnread(fields.Raw):
     def output(self, key, feed):
         if not current_user.is_authenticated():
-            return 0
+            return True
+
+        try:
+            unread = UserFeed.query.filter_by(user=current_user, feed_id=feed.id).one().unread
+        except:
```
alexgerstein: Specify except here too

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -14,13 +15,21 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
     },
 
     markAllAsRead: function() {
+        this.stopListening(this.collection, 'change:unread');
```
alexgerstein: Interesting. Didn't know this existed.

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -14,13 +15,21 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
...
+        this.model.save({unread: false});
```
alexgerstein: Do we need these two lines? If someone clicks "markAllAsRead" multiple times, it's still going to retrigger the api call because of the save.

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -55,6 +64,9 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
         });
 
         this.itemViews = itemCards;
+        if (this.itemViews.length > 0) {
```
alexgerstein: What does this do?

sourcemash/frontend/templates/sidenav.tpl
```diff
@@ -22,8 +22,8 @@
                 <li class="row collection-item">
                   <span class="col s7 truncate"><%= category.get('name') %></span>
                   <span class="badge col s2"><%=category.get('item_count')%></span>
-                  <% if (category.get('unread_count') > 0) { %>
-                    <span class="new badge col s3"><%=category.get('unread_count')%></span>
+                  <% if (category.get('unread')) { %>
+                    <span class="new badge col s3"></span>
```
alexgerstein: Should we say "New Items" or something?

alexgerstein: Or maybe just have a blue dot?

tests/factories/user_category_factories.py
```diff
@@ -0,0 +1,12 @@
...
+
+from sourcemash.models import UserCategory, Category
+from user_factories import UserFactory
+
+class UserCategoryFactory(SQLAlchemyModelFactory):
+    FACTORY_FOR = UserCategory
+    FACTORY_SESSION = db.session
+
+    user = factory.SubFactory(UserFactory)
+    category = Category("category")
```
alexgerstein: Should this be a subfactory as well?

sourcemash/api/categories.py
alexgerstein: I think we should cut it. It just adds time to API calls without serving enough functionality. It will also make the code much neater.

sourcemash/frontend/static/js/views/items_view.js
swglad: Makes sure you only update the read status if the collection is nonempty. The problem is that this.collection isn’t always full when rendered gets called, and an empty collection has no notion of read status.

sourcemash/frontend/templates/sidenav.tpl
swglad: It says “New”

alexgerstein:
Freaking awesome work! I made a few comments on ways we can clean-up the code a little, but look tremendous overall.

sourcemash/api/categories.py
swglad: Done.

sourcemash/frontend/static/js/views/items_view.js
swglad: We need to save since we need the database to sync against the UserFeed/UserCategory table now. Maybe the better way to do it is disable the button or make it only listenOnce to the event (Backbone has that built in)

alexgerstein: Won't the updateReadStatus get triggered from the for-loop where you do the save on the unread items?

sourcemash/frontend/static/js/views/items_view.js
alexgerstein: If we only update change "unread" when an item is read in the collection, then do we need to also do this in the render at all?

sourcemash/frontend/static/js/views/items_view.js
swglad: Yes, but there's no point for it to get hit 100 times for every item in a feed. Instead, I have it stopListening and just call it manually here (which is better I think).

sourcemash/frontend/static/js/views/items_view.js
swglad: Yes, because now that there's no item_counts, you can't live update the notion of "New" for a category when clicking on items in a feed or for a feed when clicking on items in a category. This makes it such that when you open a new itemsView, it checks if all the items are read and updates it accordingly.

tests/factories/user_category_factories.py
swglad: We don't have a category factory and there isn't really a use for one (since we have itemWithCategory), so it just makes a Category here for this.

swglad: But it does not make sense for the UserCategory factory to make an itemWithCategory just to get a category.

sourcemash/frontend/templates/feed.tpl
```diff
@@ -1,6 +1,7 @@
 <div class="container">
   <div class="row valign-wrapper">
-      <h2 class="col s8 m9"><%= model.get('title') %></h2>
+      <h2 class="col s6 m9"><%= model.get('title') %></h2>
+      <span class="col s2 grey-text text-lighten-1">[<%=model.get('item_count')%> items]</span>
```
swglad: Let me know if you like this.

sourcemash/api/categories.py
```diff
@@ -1,11 +1,23 @@
...
 from sqlalchemy import func
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
+MIN_ITEM_COUNT = 1
+
+class isUnread(fields.Raw):
+    def output(self, key, category):
+        try:
+            unread = UserCategory.query.filter_by(user=current_user, category_id=category.id).one().unread
+        except (MultipleResultsFound, NoResultFound):
```
alexgerstein: Should MultipleResultsFound ever happen?

sourcemash/api/categories.py
```diff
@@ -13,18 +25,41 @@
...
+    def put(self, id):
+        args = self.reqparse.parse_args()
+
+        category = Category.query.get_or_404(id)
+
+        # Mark category as Read
+        if args.unread != None:
+            try:
+                user_category = UserCategory.query.filter_by(user=current_user, category_id=category.id).one()
+            except (MultipleResultsFound, NoResultFound):
```
alexgerstein: Multiple?

sourcemash/api/categories.py
```diff
@@ -44,24 +79,13 @@ def get(self):
...
-
-            category.item_count = count
-            category.unread_count = category.item_count - read_item_count
 
-            unsubscribed_item = Item.query.filter(Item.cats.contains(category)) \
+            if count > MIN_ITEM_COUNT:
+                unsubscribed_item = Item.query.filter(Item.cats.contains(category)) \
                                           .filter(~Item.feed_id.in_(user_feed_ids)) \
                                           .first()
-            if unsubscribed_item:
```
alexgerstein: How are we handling unsubscribed items for the unread boolean?

sourcemash/api/categories.py
```diff
@@ -44,24 +79,13 @@ def get(self):
...
-                                .filter(Item.cats.contains(category)) \
-                                .filter(~UserItem.unread) \
-                                .count()
-
-            category.item_count = count
-            category.unread_count = category.item_count - read_item_count
 
-            unsubscribed_item = Item.query.filter(Item.cats.contains(category)) \
+            if count > MIN_ITEM_COUNT:
+                unsubscribed_item = Item.query.filter(Item.cats.contains(category)) \
```
alexgerstein: Do we even need this anymore?

sourcemash/api/feeds.py
```diff
@@ -27,21 +28,17 @@ def output(self, key, feed):
...
-class getUnreadCount(fields.Raw):
+class isUnread(fields.Raw):
     def output(self, key, feed):
         if not current_user.is_authenticated():
-            return 0
+            return True
+
+        try:
+            unread = UserFeed.query.filter_by(user=current_user, feed_id=feed.id).one().unread
+        except (MultipleResultsFound, NoResultFound):
```
alexgerstein: Multiple?

sourcemash/api/feeds.py
```diff
@@ -96,7 +93,7 @@ def post(self):
         # Get or Create Feed
         try:
             feed = Feed.query.filter(Feed.url==rss_feed['url']).one()
-        except:
+        except (MultipleResultsFound, NoResultFound):
```
alexgerstein: Multiple.

sourcemash/api/feeds.py
```diff
@@ -120,7 +117,7 @@ def post(self):
         try:
             subscription = current_user.subscribed.filter(Feed.id==feed.id).one()
             return {"errors": {"url": ["Already subscribed"]}}, 409
-        except:
+        except (MultipleResultsFound, NoResultFound):
```
alexgerstein: Multiple?

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -55,6 +66,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
         });
 
         this.itemViews = itemCards;
+        this.updateReadStatus();
```
alexgerstein: Why is this here again?

sourcemash/frontend/templates/feed.tpl
alexgerstein: If this is the only time we use item_count, I'm not so sure it's worth it.

tests/factories/user_category_factories.py
alexgerstein: What's weird about categories made this way in the factory is that they won't actually be present in the database i don't think. So if you wanted to access category in a test, you'll get an error saying it's not bound to the session or something like that

sourcemash/api/categories.py
swglad: No, it shouldn't. But if there is a db problem and it happens, we should catch it and return the "default" behavior as opposed to dumping a stacktrace.

swglad: Nope. Will remove.

sourcemash/api/categories.py
swglad: I was thinking about it, but I still like the feature. I think ideally it should be a setting that users turn on & off (just a column on the user table), but it's worth keeping.

sourcemash/frontend/static/js/views/items_view.js
swglad: Let's say you read every item in every feed. Each category will still be marked as "New" in the sidebar since we're not tracking read/unread by counting items anymore. So, this makes sure that when you click on a Category that says "New" with all read items in it, it updates the read status and marks it as read. 

We need this because, otherwise, there is no change on "unread" so the category will continue to be marked as unread even if every item is read since there's no change.

sourcemash/frontend/templates/feed.tpl
swglad: It still seems weird not to list the number of items anywhere. We can get away with it at the category level since there are never that many items in a category, but feeds range from 10 to 90 items so it's just kinda odd not to report that.

tests/factories/user_category_factories.py
swglad: Not an issue with tests and seems to work as expected...

sourcemash/api/categories.py
swglad: Not a problem since the PUT is triggered from the frontend so it just sees it as another item in the collection.

worker/scraper.py
```diff
@@ -105,6 +109,11 @@ def _store_items(feed):
         db.session.add(new_entry)
         db.session.commit()
 
+        new_item_count += 1
+
+    feed.item_count += new_item_count
+    db.session.commit()
+
```
alexgerstein: Need to retrigger unread = True after scraping.

swglad: Merge in your scraper split first and put UserFeed at the end of feeds and UserCategory at the end of category

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -3,6 +3,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
         if (this.model) {
             this.listenTo(this.model, 'change:subscribed change:title', this.render);
         };
+        this.listenTo(this.collection, 'change:unread', this.updateReadStatus);
```
alexgerstein: Try adding sync here.

swglad: Worked perfectly. Great call.

worker/scraper.py
swglad: Done.

worker/scraper.py
```diff
@@ -20,6 +21,11 @@
 def scrape_feed_articles(feed):
     _store_items(feed)
 
+    # Mark feed as unread for all users
+    for user_feed in UserFeed.query.filter_by(feed_id=feed.id).all():
```
alexgerstein: What if no new items came in on the _store_items call for the feed?

worker/scraper.py
```diff
@@ -49,6 +55,15 @@ def categorize_feed_articles(feed, categorizer):
         for category in categories:
             item.categories.append(category)
 
+            # Mark category as unread for all users
+            try:
+                category_model = Category.query.filter_by(category=category).one()
+                for user_category in UserCategory.query.filter_by(category=category_model).all():
```
alexgerstein: Just learned about: http://docs.sqlalchemy.org/en/latest/core/dml.html

Want to give that a shot? Looks nifty!

worker/scraper.py
swglad: Great catch. I'll have store items return a book new_items

worker/scraper.py
swglad: Not working for now but let's try it again another time.

worker/scraper.py
swglad: Actually, see directly below.

sourcemash/api/feeds.py
```diff
@@ -55,8 +52,7 @@ def output(self, key, feed):
 }
 
 feed_status_fields = {
-    'item_count': getItemCount,
-    'unread_count': getUnreadCount
+    'unread': isUnread
```
alexgerstein: Move this into feed_fields too?

sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -15,12 +16,24 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
...
+            mixpanel.track("Marked all as read");
+            unread.forEach(function(model) {
+                model.save({unread: false});
+                mixpanel.people.increment("items read");
+            });
+        };
+    },
+
+    updateReadStatus: function() {
+        if (this.collection.length > 0) {
```
alexgerstein: Do we need this if check if it only happens after the feed is fetched/synced?

sourcemash/api/feeds.py
swglad: Kk

sourcemash/frontend/templates/sidenav.tpl
```diff
@@ -41,10 +38,9 @@
           <div class="loading center-align"></div>
           <% feeds.where({subscribed: true}).forEach(function(feed) { %>
             <a href="/#feeds/<%= feed.get('id') %>"><li class="row collection-item">
-            <span class="col s7 truncate"><%= feed.get('title') %></span>
-            <span class="badge col s2"><%=feed.get('item_count')%></span>
-            <% if (feed.get('unread_count') > 0) { %>
-              <span class="new badge col s3"><%=feed.get('unread_count')%></span>
+            <span class="col s9 truncate"><%= feed.get('title') %></span>
+            <% if (feed.get('unread')) { %>
+              <span class="new badge col s3"></span>
```
alexgerstein: Instead of the heavy new label, what if we have a dot made with icon: mdi-image-brightness-1. We can color it whatever fits our theme and give it a tool tip that says "New Items"? "new" probably isn't descriptive enough because a user wouldn't know if that means it has new items or if it's a new category.

tests/factories/user_category_factories.py
alexgerstein: It may become an issue down the line, but we can leave it for now.

worker/scraper.py
alexgerstein: Got it. It's even simpler than we thought:

```python
UserCategory.query.filter_by(category=category_model).update({"unread": True})
db.session.commit()
```

worker/scraper.py
alexgerstein:
```python
UserFeed.query.filter_by(feed_id=feed.id).update({"unread": True})
db.session.commit()
```

swglad: Done. See below.

sourcemash/frontend/static/scss/sourcemash.scss
```diff
@@ -184,6 +184,12 @@ div.side-nav {
   margin: 0;
 }
 
+.new-items-badge {
+  i {
+    color: #D32F2F;
```
alexgerstein: Can we use the @extend functionality here?

PR #297, opened 2015-05-15
Appropriate links open in new page

### LINKED ISSUE #279: Open links in new tab

swglad:
Resolves #279.

### DIFF
--- a/sourcemash/frontend/static/js/views/items_view.js
+++ b/sourcemash/frontend/static/js/views/items_view.js
@@ -52,6 +52,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
 
             var itemCardViewOneCol = new Sourcemash.Views.ItemCardView({el: "#item-" + item.get('id') + "-onecol", model: item, user: user });
             itemCards.push(itemCardViewOneCol);
+            $(".card-reveal a").attr("target","_blank");
         });
 
         this.itemViews = itemCards;
--- a/sourcemash/frontend/templates/footer.tpl
+++ b/sourcemash/frontend/templates/footer.tpl
@@ -3,14 +3,14 @@
     <div class="row">
       <div class="col s12">
         <h5 class="white-text">Help Sourcemash Grow</h5>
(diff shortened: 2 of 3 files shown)


sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -52,6 +52,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
 
             var itemCardViewOneCol = new Sourcemash.Views.ItemCardView({el: "#item-" + item.get('id') + "-onecol", model: item, user: user });
             itemCards.push(itemCardViewOneCol);
+            $(".card-reveal a").attr("target","_blank");
```
alexgerstein: Can we not put this into the tpl directly, rather than using jquery?

swglad: But we want to apply this only to the contents of the card reveal, which we won’t know what it is since we scrape it...

swglad: So, no we can't.

alexgerstein: Why can't we put it in item-card.tpl.

swglad: In a <script> tag?

alexgerstein: Why does it have to be a script tag? I may be missing something, but can't you just do
``` HTML
<p class="flow-text"><a href="<%= item.get('link') %>" target="_blank"><%= item.get('title') %></a></p>
```

swglad: Yes, you are missing something. The problem is that we render item.summary as HTML and we want THOSE links to open in new tabs as well. So we need to either process that using beautiful soup on the backend (nontrivial - I tried it) or jquery on the frontend (the one line you see here).

alexgerstein: Ooooohhhh. That makes sense. Thanks for the clarification. Looks good then!

alexgerstein:
Let's move the jquery links to the item-card view.

PR #298, opened 2015-05-15
Tooltip for mark as read

### LINKED ISSUE #280: Tooltip for "Mark as Read"

swglad:
Resolves #280.

### DIFF
--- a/sourcemash/frontend/static/js/views/items_view.js
+++ b/sourcemash/frontend/static/js/views/items_view.js
@@ -28,6 +28,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
 
         // Render parent view
         this.$el.html(this.template({ model: this.model, items: this.collection.models }));
+        this.$('.tooltipped').tooltip({delay: 50});
 
         // Render subscribe modal view
         this.subscribeModalView = new Sourcemash.Views.SubscribeModalView({ collection: this.collection });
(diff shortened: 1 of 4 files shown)


sourcemash/frontend/static/js/views/items_view.js
```diff
@@ -28,6 +28,7 @@ Sourcemash.Views.ItemsView = Backbone.View.extend({
 
         // Render parent view
         this.$el.html(this.template({ model: this.model, items: this.collection.models }));
+        this.$('.tooltipped').tooltip({delay: 50});
```
alexgerstein: We shouldn't need this line, because as the Materialize docs explain: "Tooltips are initialized automatically"

sourcemash/frontend/templates/category.tpl
```diff
@@ -2,7 +2,7 @@
   <div class="row valign-wrapper">
       <h2 class="col s11"><%= model.get('name') %></h2>
 
-      <span title="Mark as Read" class="mark-all-read col s1"><i class="medium mdi-action-done-all"></i></span>
+      <span title="Mark as Read" class="mark-all-read col s1"><i class="medium mdi-action-done-all tooltipped" data-position="bottom" data-tooltip="Mark All as Read"></i></span>
```
alexgerstein: All uppercased on Mark All As Read? Your call.

swglad: I looked at it both ways but wasn’t too committed to either.

sourcemash/frontend/static/js/views/items_view.js
swglad: Not true based on testing just now.

PR #299, opened 2015-05-15
Update README.md

### LINKED ISSUE #285: Update readme
Doesn't contain ```redis-server &; python manage.py worker``` to get the queue going.

swglad:
Closes #285.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,10 +28,10 @@ The site was built for the CS98 course at Dartmouth College.
 
 ## Run Locally
 1. Start up a local server: ```python manage.py server```
-2. In another terminal session,
+2. In another terminal session, [optional]
 	1. Start up a redis server: ```redis-server &```
 	2. Start up an instance of the worker: ```python manage.py worker &```
-3. Start up the scraper: ```python manage.py scrape```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,11 +28,14 @@ The site was built for the CS98 course at Dartmouth College.
 
 ## Run Locally
 1. Start up a local server: ```python manage.py server```
-2. Start up the scraper: ```python manage.py scrape```
+2. In another terminal session,
```
alexgerstein: Should we explain that these are optional?

swglad: Do you think we should explain the noqueue thing and this? How do you think we should do that?

swglad: I think we should leave it like this because this is the (theoretically) ideal way to use the app locally.

alexgerstein: I don't think it really matters how we write out the readme because no one will be using it other than us. However, I find it's a pain personally to set up the queueing and everything locally. Especially since the tasks put into the queue don't go away when you shut down the server. When you turn the server back on, the events continue. My "ideal" way to use the app is with a nonlocal scrape once. It makes it easier.

PR #306, opened 2015-05-17
Sync categories when feeds are subscribe toggled

### LINKED ISSUE #275: Say "refresh" in category tab

swglad:
Resolves #275.

### DIFF
--- a/sourcemash/frontend/static/js/views/sidenav_view.js
+++ b/sourcemash/frontend/static/js/views/sidenav_view.js
@@ -6,7 +6,8 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
     this.feeds = options.feeds;
     this.categories = options.categories;
     this.loading = true;
-    this.listenTo(this.feeds, 'sync change:subscribed change:unread', this.render);
+    this.listenTo(this.feeds, 'sync change:unread', this.render);
+    this.listenTo(this.feeds, 'change:subscribed', function(){ this.categories.fetch(); });
     this.listenTo(this.categories, 'sync change:unread', this.render);
   },
 


sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -6,7 +6,8 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
     this.feeds = options.feeds;
     this.categories = options.categories;
     this.loading = true;
-    this.listenTo(this.feeds, 'sync change:subscribed change:unread', this.render);
+    this.listenTo(this.feeds, 'sync change:unread', this.render);
+    this.listenTo(this.feeds, 'change:subscribed', function(){ this.categories.fetch(); });
```
swglad: So this line works perfectly as expected when adding (subscribing to) feeds: more categories appear and everything is re-rendered as desired.

However, it doesn't work quite as well when you unsubscribe/remove feeds. Then, we get some residual categories that only have one item and don't disappear. This makes sense because we don't remove categories from this.categories and fetch doesn't do a reset-and-refill. 

Generally, I'd say this is not that big of an issue because the likelihood of highly repetitive subscribes and unsubscribes is pretty low, and it's not the worst thing if some old categories hang around until you refresh the page.

alexgerstein: What if you do this.categories.reset(this.categories)?

swglad: It makes the render jump / unclean

alexgerstein: Ahh. Then yeah, it's probably fine just to leave it.

PR #316, opened 2015-05-19
Refresh button on feed subscribe change for categories

### LINKED ISSUE #315: Fetch categories bug
The listenTo from #306 doesn't work right. We should just add a "Refresh" button at the top of the categories list instead of doing this.categories.fetch().

swglad:
Resolves #315.

### DIFF
--- a/sourcemash/frontend/static/js/views/sidenav_view.js
+++ b/sourcemash/frontend/static/js/views/sidenav_view.js
@@ -6,8 +6,16 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
     this.feeds = options.feeds;
     this.categories = options.categories;
     this.loading = true;
+    this.feedsChanged = false;
+
     this.listenTo(this.feeds, 'sync change:unread', this.render);
-    this.listenTo(this.feeds, 'change:subscribed', function(){ this.categories.fetch(); });
+    this.listenTo(this.feeds, 'change:subscribed',
+      function(){
+        if (!this.loading) {
+          this.feedsChanged = true;
+          this.render();
+        };
+    });
     this.listenTo(this.categories, 'sync change:unread', this.render);
   },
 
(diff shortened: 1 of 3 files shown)


sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -61,13 +71,18 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
     });
   },
 
+  refreshPage: function(e){
+    e.preventDefault();
+    window.location.reload(true);
```
alexgerstein: Let's just fetch/reset the categories again instead of reloading the whole page.

alexgerstein: http://backbonejs.org/#Collection-reset

swglad: Every iteration of this I've been playing with hasn't been successful. Not sure what to do at this point. I think we should just bite the bullet on the reset, knowing that it actually works correctly and nothing in the reset-and-fetch variety seems to.

alexgerstein:
Okay. We can merge this in and maybe I'll try to play around with it later.

sourcemash/frontend/static/js/views/sidenav_view.js
```diff
@@ -6,8 +6,17 @@ Sourcemash.Views.SidenavView = Backbone.View.extend({
...
     this.categories = options.categories;
     this.loading = true;
+    this.feedsChanged = false;
+
     this.listenTo(this.feeds, 'sync change:unread', this.render);
-    this.listenTo(this.feeds, 'change:subscribed', function(){ this.categories.fetch(); });
+    this.listenTo(this.feeds, 'change:subscribed',
+      function(){
+        // Don't tell the user to refresh categories until fetch is complete
+        if (!this.loading) {
```
alexgerstein: Do we still need this?

swglad: Yes. First, because we don't want both a refresh button and a loading wheel appearing simultaneously. It looks like we don't know what to tell the user. It should be "act" *or* "wait", not act *and* wait.

Issue #231, opened 2015-04-27
Pagination.


alexgerstein:

swglad:
Probably worth prioritizing #212 Prevent (obvious) vandalism if this looks like it'll be too much work. Writing the validation in the feeds API is more important.

swglad:
Lowest priority of Week 15 frontend issues.

PR #326, opened 2015-06-07
Add limit on feed items.

alexgerstein:
Limit the number of items sent to a feed to speed up load time. In the future, it would be best to add pagination like described in #231.

### DIFF
--- a/sourcemash/api/items.py
+++ b/sourcemash/api/items.py
@@ -154,7 +154,9 @@ class FeedItemListAPI(Resource):
     def get(self, feed_id):
         return {'items': [marshal(item, item_fields)
                 for item in Item.query.filter_by(feed_id=feed_id)
-                                      .limit(MAX_ARTICLES).all()]}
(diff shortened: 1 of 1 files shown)


sourcemash/api/items.py
```diff
@@ -126,13 +145,17 @@ def get(self):
...
+                for id, count in trending_items]}
 
 
 class FeedItemListAPI(Resource):
 
     def get(self, feed_id):
-        return {'items': [marshal(item, item_fields) for item in Item.query.filter_by(feed_id=feed_id).all()]}
+        return {'items': [marshal(item, item_fields)
+                for item in Item.query.filter_by(feed_id=feed_id)
+                                      .limit(MAX_ARTICLES).all()]}
```
swglad: sort by last_updated?

alexgerstein: It probably does this by default since ids increase, but i made the change just to be sure.

PR #327, opened 2015-06-07
Check if item title exists

swglad:
Before committing and creating a new item.


alexgerstein:
If it works, we should do it with feed_id instead of this....

swglad:
Bad solution.

PR #328, opened 2015-06-07
Fix duplicate items bug

swglad:
Objects committed to a queue keep their CURRENT STATE regardless of whether they are changed in the db, so you MUST query new instances of these objects from the database in queued tasks.


alexgerstein:
Just missing the call in feeds POST. Then it should be good to go.

